Issue #144, opened 2023-02-18
Forcing a specific version of jax and jaxlib


EiffL:
I see that the requirements are forcing version 0.3 of jax, is there a good reason for it?

https://github.com/astro-informatics/s2fft/blob/e9e6f78c685d2dd54f6344e29a35de65c09fb350/requirements/requirements-core.txt#L7

In general, you don't want to impose a jax version, as it can cause mannnny compatibility problems. It's ok to impose a minimal version though, like "jax>=0.3.1"

Also, it's good practice not to explicitly impose a requirement on jaxlib, because you typically need to be careful about how you install jaxlib for a particular system and if pip just tries to download a random version it may uninstall a good jaxlib to install a bad jaxlib.

CosmoMatt:
Do you think it would be best to have jax and jaxlib as open requirements then @EiffL ? e.g. remove all versions from requirements-core.txt? If so this should take no time to fix.

EiffL:
A good point of reference is to look at one of Google's own jax packages, like this one:
https://github.com/google/flax/blob/main/setup.py
They only have `jax` as the requirement, and it is ok to add the version number. Because it is very possible that previous version of jax were not working in the same way as the new ones... so the one you currently have is good I think.

jasonmcewen:
Can we keep this pinned but upgrade to 0.4.x?

EiffL:
well the issue is forcing a specific version of jax. JAX is getting a new release like every 2 weeks. If all jax-based libraries imposed a specific version, it would be utterly unmanageable. 

The idea is that all libraries should have a minimum version under which they are not guaranteed to work, and then try to remain up to date with upstream jax. breaking changes are rare these days.

EiffL:
One of the core strengths of jax is all libraries based on it are interoperable. You can grab your favorite optim library (optax, jaxopt, etc) combine it with your favorite NN library (haiku, flax, equinox, etc.). It would be impossible for all of them to agree on just one version of jax to follow.

jasonmcewen:
I see. Ok, let's set a minimum then.

CosmoMatt:
Closing as this was addressed in #145

Issue #143, opened 2023-02-18
Forcing jax float64 mode


EiffL:
This is not really good practice, users can have legitimate reasons for not using float64 and if they import `s2fft` first, it will force the type without leaving them a choice:
https://github.com/astro-informatics/s2fft/blob/e9e6f78c685d2dd54f6344e29a35de65c09fb350/s2fft/transforms/spherical.py#L3

If you think float64 is really important for this library, you could instead send a warning back if the users have not enabled it.

CosmoMatt:
The problem really starts to become apparent inside the recursions. With 64bit precision, at high L the error can grow from 1e-16 to around 1e-13 or so. With 32bit precision, without accounting for any other factors, this should correspond to an error from around 1e-6 to 1e-3 to even 1e-2, which isn't great.

I think a warning would be a good option around this though, but we will have to be clear that 64bit can be really quite important in this case. This hard coding of 64bit precision comes up throughout several functions, perhaps we should just include a warning at the high-level?

EiffL:
Yep, I'm proposing a solution in #145 Maybe the warning can be evn more ominous ^^

CosmoMatt:
Closing as this was addressed in #145

PR #145, opened 2023-02-18
Suggested little fixes (jax version, float64)

EiffL:
I'm opening this PR to suggest a few fixes as I see them, in particular addressing #143  #144.

Still at draft stage, because I want to check if everything works ok in the testing suite

### DIFF
--- a/notebooks/spherical_harmonic_transform.ipynb
+++ b/notebooks/spherical_harmonic_transform.ipynb
@@ -28,6 +28,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "from jax.config import config\n",
+    "config.update(\"jax_enable_x64\", True)\n",
+    "\n",
     "import numpy as np\n",
     "import s2fft \n",
     "\n",
@@ -164,7 +167,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3.9.0 ('s2fft')",
+   "display_name": "Python 3.8.16 64-bit ('s2fft')",
    "language": "python",
    "name": "python3"
   },
@@ -178,12 +181,12 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.0"
+   "version": "3.8.16"
   },
   "orig_nbformat": 4,
   "vscode": {
    "interpreter": {
-    "hash": "3425e24474cbe920550266ea26b478634978cc419579f9dbcf479231067df6a3"
+    "hash": "d6019e21eb0d27eebd69283f1089b8b605b46cb058a452b887458f3af7017e46"
    }
   }
  },
--- a/notebooks/wigner_transform.ipynb
+++ b/notebooks/wigner_transform.ipynb
@@ -28,6 +28,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "from jax.config import config\n",
+    "config.update(\"jax_enable_x64\", True) \n",
+    "\n",
     "import numpy as np\n",
     "import s2fft \n",
     "\n",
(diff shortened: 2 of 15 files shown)


CosmoMatt:
Looks great, I've activated CI for you @EiffL, which will take some time to run - I'll also check on a GPU machine to check precision there. Then we should be able to merge. Thanks :)

CosmoMatt:
@EiffL CI failing due to no jaxlib version. Two options, either we add a minimal requirement to requirements-core.txt or we manually install jaxlib during the tests.yml workflow. 

In practise some users may not want to manage their own version of jaxlib, so if we can include a minimal version that would be ideal I think?

EiffL:
I think if we do it like this it should work, it would grab the current jaxlib version.

As a bonus, now s2fft works on  colab :-) 
https://colab.research.google.com/drive/12LgYB6sfljd4YYO9ifGAnCTPLEv7QOot?usp=sharing

CosmoMatt:
Very cool, I spent a few minutes trying to get something nice set up with colab but GPU wasn't playing ball and I had other urgent work. Great this works with this update, we should add the notebook as a badge in the readme. I still need to check the change to 64bit precision on GPU but haven't had a chance yet

CosmoMatt:
@EiffL Ok I've run some basic transforms to check 64bit precision
- [ ] without specifying 64bit, correctly defaults to 32bit
- [ ] without specifying 64bit, roundtrip error at L = 128 with MW (which should be exact) is ~1e-2
- [ ] with specifying 64bit, roundtrip error at L = 128 with MW (which should be exact) is 1e-14

Basically exactly what I predicted which is nice :sweat_smile: Otherwise though this looks like its doing some nice functionality, should pull into main. We should probably make the warning even more severe though, because 1e-2 error at 128 indicates that for larger L things could get really quite bad.

EiffL:
Yep ok, feel free to suggest a more appropriate warning ^^

CosmoMatt (approved):
@EiffL Once CI's have passed we should be good to merge into main, thanks for this! Once in main we can link colab notebook into the primary readme.

jasonmcewen:
I'd just like to review this before merging to main to keep abreast of revisions. Will take a look now...

jasonmcewen (approved):

jasonmcewen:
Looks good!

jasonmcewen:
Oh, do we need to update the notebooks to force double precision?

jasonmcewen:
@EiffL Btw, in your notebook (https://colab.research.google.com/drive/12LgYB6sfljd4YYO9ifGAnCTPLEv7QOot?usp=sharing), if you set the healpy transforms to not do any iterations we should get identical results.

jasonmcewen:
Oh, one other point @EiffL.  Our L is different to healpix lmax, with L=lmax+1.

EiffL:
yep I saw that :-) I'm just using that notebook to experiment with improving the jitting time of the healpix transforms, but I think your existing notebooks are better. 
Precviously they woudlnt be able to use the GPU on colab because installing s2fft would uninstall the cuda-enable jaxlib ^^

CosmoMatt:
> 

This is a good point, we should just add the config lines at the top of the two notebooks before merging.

jasonmcewen:
@EiffL @CosmoMatt Is this good to merge now?  Does someone want to push the button? 

Our general process up until now has been for the person who submitted the PR to do the final merge once approved.  However, I guess that's not such a typical process now the repo is public, i.e. you're probably both expecting the other to do the merge!

CosmoMatt:
Should be fine to merge @jasonmcewen, if you're happy with the changes to your PR @EiffL we'll merge this and close the associated issues.

EiffL:
Go for it!

PR #147, opened 2023-02-19
Debug

CosmoMatt:
Fixes some floating bugs which are causing downstream issues.

- [x] Remove tox from CI as docs keep failing to link correctly (solution isn't as elegant but should hopefully be more robust).
- [x] Switch CI from py38 to py39.
- [x] Explicitly null undefined values in jax pre computations (not an issue for forward pass, but for certain transforms when executing auto-grad these undefined values are folded into the adjoint computation, causing gradients to become undefined in some cases).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


jasonmcewen (approved):
Make a minor commit just to clarify the comments about setting nans to zeros.

PR #162, opened 2023-03-13
Feature/custom grads

CosmoMatt:
This PR switches out fully automatic differentiation to a `custom_vjp` implementation. There are two options regarding this, as the spherical harmonic transform is holomorphic, the right multiplication of gradients in the [Vector-Jacobian Product (VJP)](https://jax.readthedocs.io/en/latest/notebooks/autodiff_cookbook.html) is invariant under transposition, leading to:

- VJP of forward transform = conjugate of the inverse spherical harmonic transform of the conjugate of the cotangents.
- JVP of forward transform = forward transform of tangents.

And similarly for the inverse spherical harmonic transforms. The issue with this, is that we have a resampling step for "mw" and "mwss" which slightly modifies this relationship, and would require some, potentially very ugly, code. Instead, we can strike a middle ground by explicitly defining the VJP for the latitudinal step alone, and leave the remaining  propagation of the cotangent to vanilla AD. This avoids the memory overhead associated with applying grad of a scan with large carry terms.

**tl;dr: with this update the gradients should be approximately the same cost as the transforms, both in terms of memory and wall-time (they can also be distributed over multiple GPUs on a single node with [SPMD](https://jax.readthedocs.io/en/latest/glossary.html#term-SPMD)).**

### Core Content
- [x] Add custom gradient for forward spin-spherical harmonic transform
- [x] Add custom gradient for inverse spin-spherical harmonic transform
- [x] Filter precomputed matrices into custom gradients
- [x] Ensure gradient functionality extends to Wigner transforms
- [x] Add unit tests for added functionality (all to machine precision, and $\mathcal{O}(L^2)$ peak memory on GPU device).

### Additional Content
- [x] Reactivated CI as Conan SSHT wheel issues have been resolved
- [x] Minor tidying of various, now redundant, variables.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,14 +25,14 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     
-    # - name: Install dependencies
-    #   run: |
-    #     python -m pip install --upgrade pip
-    #     pip install -r requirements/requirements-tests.txt
-    #     pip install -r requirements/requirements-core.txt
-    #     pip install .
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements/requirements-tests.txt
+        pip install -r requirements/requirements-core.txt
+        pip install .
     
-    # - name: Run tests
-    #   run: |
-    #     pytest --cov-report term --cov=s2wav --cov-config=.coveragerc 
-    #     codecov --token 298dc7ee-bb9f-4221-b31f-3576cc6cb702
+    - name: Run tests
+      run: |
+        pytest --cov-report term --cov=s2wav --cov-config=.coveragerc 
+        codecov --token 298dc7ee-bb9f-4221-b31f-3576cc6cb702
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ tox -e py38           # for tox
 ```
 
 In the very near future one will be able to install `S2FFT` directly
-from [PyPi]{.title-ref} by `pip install s2fft` but this is not yet
+from [PyPi]{https://pypi.org} by `pip install s2fft` but this is not yet
 supported. Note that to run `JAX` on NVIDIA GPUs you will need to follow
 the [guide](https://github.com/google/jax#installation) outlined by
 Google. 
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ pixels of equal areas, which has many practical advantages.
 
 The Python dependencies for the `S2FFT` package are listed in the file
 `requirements/requirements-core.txt` and will be automatically installed
-into the active python environment by [pip]{.title-ref} when running
+into the active python environment by [pip](https://pypi.org) when running
 
 ``` bash
 pip install .        
@@ -75,7 +75,7 @@ tox -e py38           # for tox
 ```
 
 In the very near future one will be able to install `S2FFT` directly
-from [PyPi]{https://pypi.org} by `pip install s2fft` but this is not yet
+from [PyPi](https://pypi.org) by `pip install s2fft` but this is not yet
 supported. Note that to run `JAX` on NVIDIA GPUs you will need to follow
 the [guide](https://github.com/google/jax#installation) outlined by
 Google. 
--- a/notebooks/custom_gradients.ipynb
+++ b/notebooks/custom_gradients.ipynb
@@ -1,0 +1,112 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "cpu\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Specify CUDA device\n",
+    "import os\n",
+    "os.environ['CUDA_VISIBLE_DEVICES'] = ''\n",
+    "os.environ['JAX_CHECK_TRACER_LEAKS'] = 'True'\n",
+    "\n",
+    "from jax.config import config\n",
+    "config.update(\"jax_enable_x64\", True)\n",
+    "\n",
+    "# Check we're running on GPU\n",
+    "from jax.lib import xla_bridge\n",
+    "print(xla_bridge.get_backend().platform)\n",
+    "\n",
+    "import jax\n",
+    "from jax import jit, grad \n",
+    "import jax.numpy as jnp \n",
+    "from jax.test_util import check_grads\n",
+    "import numpy as np \n",
+    "\n",
+    "import s2fft "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "L = 16\n",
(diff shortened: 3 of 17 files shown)


CosmoMatt:
## Original issue:

Running auto-grad over the transform raised $\mathcal{O}(L^3)$ memory overhead, which is prohibitive. This was coming from running grad over the recursions lax.fori_loop which is implemented in terms of a scan. With vanilla auto-grad JAX preallocates all memory associated with the scan which is the carry $\mathcal{O}(L^2)$ times the number of iterations $\mathcal{O}(L)$, hence $L^3$. See e.g. this [issue](https://github.com/google/jax/discussions/10131) for a somewhat similar scenario.

## Solution:

@jasonmcewen below is an abridged version of the rational for what is implemented, which should hopefully make the code easier to follow. In the end the update to the code was (thankfully) fairly minor.

#### Harmonic transform

Consider the space of square integrable functions on the sphere $\text{L}^2(\mathbb{S}^2)$, with inner product of $f$ with spin-spherical harmonic basis functions given by

$$
  {}_sf_{\ell m}
  \equiv \Big\langle {}_sf(\theta, \phi), {}_sY_{\ell m}(\theta, \phi) \Big\rangle
  = \int_{\mathbb{S}^2} \text{d}\Omega(\theta, \phi) {}_sf(\theta, \phi) {}_sY_{\ell m}^*(\theta, \phi),
$$

where $\text{d}\Omega(\theta, \phi) = \sin(\theta)\text{d}\theta\text{d}\phi$ is the standard Haar measure on $\mathbb{S}^2$. A discrete approximation to this continuous integral over the sphere is given by

$$
  {}_sf_{\ell m}
  = \sum_{\theta \phi} q(\theta) {}_sf(\theta, \phi) {}_sY_{\ell m}^*(\theta, \phi),
$$

where $q(\theta)$ are quadrature weights associated with the sample points $\theta$ and $\phi$, from which the integral can be evaluated exactly. Conversely, the spherical signal ${}_sf(\theta, \phi)$ may be synthesised from its harmonic coefficients by

$$
  {}_sf(\theta, \phi)
  = \sum_{\ell m} {}_sf_{\ell m} {}_sY_{\ell m}(\theta, \phi).
$$

####  Custom Gradients (General)

To compute the JVP associated with the forward spin-spherical harmonic transform we require a function for the pushforward $(d\varphi_x)_{\ell m}^{\theta \phi}$ of tangent vectors $v(\theta, \phi)$ into the tangent space of the codomain of $\varphi$. That is, we require a mapping characterised by the Jacobian matrix which can be written as

$$
  (d\varphi_x)_{\ell m}^{\theta \phi}
  = \frac{\partial {}_sf(\theta, \phi)}{\partial {}_sf_{\ell m}}
  = q(\theta) {}_sY_{\ell m}^*(\theta, \phi)
  = J_{\ell m}^{\theta \phi}.
$$

Fortunately our Jacobian is independent from $x$, and the JVP of the forward spherical harmonic transform is unambiguously given by

$$
  \tilde{v}_{\text{JVP}}
  = \partial \varphi_x \circ v
  = J_{\ell m}^{\theta \phi}v_{\theta \phi}
  = \sum_{\theta \phi} q(\theta) v(\theta, \phi) {}_sY_{\ell m}^*(\theta, \phi) = \tilde{v}_{\ell m},
$$

which is simply the forward spin-spherical harmonic transform of the tangents, resulting in output tangent vectors $\tilde{v}_{\ell m}$ in harmonic space. Given that the spin-spherical harmonic transform is holomorphic and recalling that the linear component of the VJP for a holomorphic function is invariant under transposition, we can write the VJP of the forward transform as

$$
  \tilde{v}_{\text{VJP}}
  = v^T \circ \partial \varphi_x
  = \partial \varphi_x^T \circ v
  = J^{\ell m}_{\theta \phi}v_{\ell m}
  = q(\theta) \sum_{\ell m} v_{\ell m} {}_sY_{\ell m}^*(\theta, \phi) = \tilde{v}(\theta, \phi),
$$

from which it is apparent that

$$
  \sum_{\ell m} v_{\ell m} {}_sY_{\ell m}^*(\theta, \phi)
  = \Bigg( \sum_{\ell m}  v^{\ast}_{\ell m} {}_sY_{\ell m}(\theta, \phi)  \Bigg)^{\ast},
$$

which is the complex conjugate of the inverse spherical harmonic transform of the conjugate of the cotangent $v_{\ell m}$. Hence, the VJP of the forward spherical harmonic transform is given by the quadrature weighted conjugate of the inverse spherical harmonic transform of the conjugate of the cotangent. Corresponding expressions for the JVP and VJP of the inverse spin-spherical harmonic transform can be derived, which are identical up to the absence of quadrature weighting.

####  Custom Gradients (Hybrid)

In fact, given the modular way in which the code is written we can have a middle ground between manual and automatic gradients, at least in this case. The problematic component of the transform can be seen from the following. Recall the forward spherical harmonic transform

$$
  {}_sf_{\ell m}
  \equiv \Big\langle {}_sf(\theta, \phi), {}_sY_{\ell m}(\theta, \phi) \Big\rangle
  = \int_{\mathbb{S}^2} \text{d}\Omega(\theta, \phi) {}_sf(\theta, \phi) {}_sY_{\ell m}^*(\theta, \phi),
$$

and noting the relations 

$$
{}_sY_{\ell m}(\theta, \phi) = (-1)^s\sqrt{\frac{2\ell+1}{4\pi}} D_{m,-n}^{\ell *} (\varphi, \theta, 0)
\quad \text{and} \quad
D_{mn}^{\ell} (\alpha, \beta, \gamma) = e^{-im\alpha} d_{mn}^{\ell}(\beta)e^{-in\gamma}
$$

from which we can rewrite the forward transform as 

$$
{}_sf_{\ell m} = (-1)^{-s}\sqrt{\frac{2\ell+1}{4\pi}} \sum_{\theta \varphi} q(\theta) {}_sf(\theta, \varphi) e^{im\varphi} d^{\ell}_{m, -s}(\theta)
$$

which, ignoring constant terms, is just a Fast Fourier transform (FFT) followed by a Hadamard product with small wigner-d (real polar d) functions. This second step is computed recursively and is exactly where the memory issue is coming from. Suppose $f_{\theta m}$ are the outputs from the FFT step, and ignoring constants we have 

$$
{}_sf_{\ell m} \propto \sum_{\theta} {}_sf_{\theta m}d^{\ell}_{m, -s}(\theta).
$$

Now the VJP for this problem is actually quite straightforward, despite $f$ being a complex valued function this mapping is trivially holomorphic, and the wigner-d kernel in this case is purely real so we don't need to worry about complex conjugation as we did above. In this case the VJP is simply given by 

$$
{}_sf_{\theta m} \propto \sum_{\ell} {}_sf_{\ell m}d^{\ell}_{m, -s}(\theta),
$$

which can be implemented in theory by just switching the axis over which we sum within our OTF recursions. In practice there are some differences between the forward and inverse recursions, again related to the resampling steps introduced for "mw" and "mwss" sampling schemes, but in the end these differences are very minor (see [here](https://github.com/astro-informatics/s2fft/blob/8c4227daf7f31cbed25ec531a7cc7e7606e18a1a/s2fft/transforms/spherical.py#L259)).

jasonmcewen (approved):

PR #170, opened 2023-11-29
Tests for consistency of HEALPix FFT and IFFT implementations

matt-graham:
Adds some tests for consistency between the NumPy and JAX implementations of HEALPix fast Fourier transform (FFT) and inverse FFT functions in `s2fft.utils.healpix_ffts`. 

Assumes that

1. `healpix_fft_numpy` and `healpix_fft_jax` should give consistent outputs (up to floating point error) for equivalent input arguments
2. `healpix_ifft_numpy` and `healpix_ifft_jax` should give consistent outputs (up to floating point error) for equivalent input arguments,
3. `healpix_fft_numpy` and `healpix_ifft_numpy` as partial functions only on first arguments are inverses of each other,
4. `healpix_fft_jax` and `healpix_ifft_jax` as partial functions only on first arguments are inverses of each other.

I'm not sure that 3 and 4 are necessarily true or not.

At the moment only the tests for 1 all pass with tests for 2, 3 and 4 failing. This may well be due to the assumptions above being invalid.

### DIFF
--- a/tests/test_healpix_ffts.py
+++ b/tests/test_healpix_ffts.py
@@ -1,0 +1,56 @@
+import numpy as np
+import pytest
+from jax import config
+from s2fft.utils.healpix_ffts import (
+    healpix_fft_jax,
+    healpix_fft_numpy,
+    healpix_ifft_jax,
+    healpix_ifft_numpy,
+)
+
+
+config.update("jax_enable_x64", True)
+
+
+@pytest.mark.parametrize("L", (32, 64))
+@pytest.mark.parametrize("nside", (4, 8, 16))
+@pytest.mark.parametrize("reality", (True, False))
+def test_healpix_fft_jax_numpy_consistency(rng, L, nside, reality):
+    f = rng.standard_normal(size=12 * nside**2)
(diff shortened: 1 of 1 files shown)


matt-graham:
@CosmoMatt has confirmed assumptions (3) and (4) above will not necessarily hold:
> FFT step for HEALPix sampling is exactly where they lose accuracy, primarily because they throw away samples in the polar regions. So we wouldn't expect to get back to the same input, other than in the equitorial band that is.
so removed the corresponding tests.

CosmoMatt:
@matt-graham thanks for setting this up. I've added a fix, basically the way the test was written you are comparing the consistency between two functions applied to an input vector. However, when the first (numpy baseline) function is applied it appears to update the input vector, so the second function is being applied to a different vector, hence producing different results. 

I've added a "fix" to the test which explicitly copies the input vector to avoid this failure mode.

matt-graham:
Should we add some tests here that check that the `healpix_*fft_*` don't mutate their arguments or do you think that would be better as a separate PR along with a fix for the current behaviour?

Issue #140, opened 2023-02-15
Explore removing loop in HEALPix FFTs (to reduce JIT compile time for high L)


jasonmcewen:

EiffL:
Yep I think that can be improved a bit

EiffL:
oookkkk I admit defeat.... I can't figure out to do it, I was hoping some clever uses of  https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.dynamic_slice.html
would work, but no, the size has to be statically known....

Of course, the nukular option remains... One could implement `healpix_fft_jax` directly in cuda, and wrap it in JAX.... It's only a matter of running a bunch of ffts so it's pretty trivial. Here is an example of how it's done: https://github.com/dfm/extending-jax

But that brings two questions:
- Would you be happy for `s2fft` to also have some custom compiled ops? The potential problem is that it would require some compilation as opposed to a pure jax library. Plus, long term maintenance has higher cost because you have to keep up with the undocumented jax custom op api.

- As long as we are implementing things in CUDA, would it be smarter to directly wrap a full SHT for healpix as opposed to just the ring-wise FFTs?

jasonmcewen:
The nuclear option is a possibility but as a very last resort I would say.  As you say, that is not going to be the easiest to maintain.  But if that is the only way we could do it.  I think implementing the minimal amount of code in cuda would be best, which would help with maintainability and shouldn't have any further performance impact, so basically just `healpix_fft_jax` etc.  But I think we should only do this as a last report.  It would be worth trying to explore some other alternatives first.  

Beyond a more efficient JAX implementation of the current HEALPix algorithm, we could perhaps consider variants of the algorithm that avoid this issue or at least mitigate it, e.g. perhaps we could compute a fixed higher resolution ring FFT and downsample rings.  Will need to look into this is further detail and give it some more thought.

CosmoMatt:
> oookkkk I admit defeat.... I can't figure out how to do it, I was hoping some clever uses of https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.dynamic_slice.html would work, but no, the size has to be statically known.... CodeGPT/ChatGPT/Copilot were no help at all :-( (there might be some use cases for good old humans after all ^^')
> 
[...]

PR #171, opened 2023-12-01
Reducing compile time of JAX HEALPix (I)FFT implementations

matt-graham:
Related to #140 though this doesn't completely remove the loops in the JAX HEALPix FFT and IFFT implementations, but it does reduce the number of unrolled operations and so compile time. Unfortunately the optimizations do make the code a bit less readable and less directly tied to the NumPy implementations.

I've tested locally against the tests added in #170 which pass, but we would probably want to merge that in first, so I'm marking this as draft until that is merged in and we can then rebase on top of that.

Compared to previous implementations, this tries to vectorize operations as much as possible by processing data connected to $\theta$ rings of the same size (all equatorial rings and the pairs of polar rings of equal sizes) together.

The big gain is in vectorizing the operations on all the equally sized equatorial bands together, as this removes around `2 * nside` unrolled loop iterations in favour of one set of vectorized operations. Processing the pairs of polar rings together gives a smaller but still helpful reduction in compile time.

### DIFF
--- a/s2fft/utils/healpix_ffts.py
+++ b/s2fft/utils/healpix_ffts.py
@@ -1,4 +1,4 @@
-from jax import jit
+from jax import jit, vmap
 
 import numpy as np
 import jax.numpy as jnp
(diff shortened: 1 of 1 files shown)


CosmoMatt (approved):
Great @matt-graham this should help for now. As you say, this approach still has the same scaling due to the outer loops but should be at least a factor of 2 less compile time, likely a little better.

jasonmcewen (approved):

PR #173, opened 2023-12-06
Switching to `pyproject.toml` for specifying package metadata

matt-graham:
Switches from current `setup.py` approach of specifying package metadata to instead use [a `pyproject.toml` configuration file](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) and sets up [`setuptools_scm`](https://github.com/pypa/setuptools_scm) to automatically set package version number from Git release tags.

I think I've directly translated all the arguments passed to `setup` function in previous `setup.py` to equivalent `pyproject.toml` properties, so this should retain same metadata when publishing to PyPI, other than there was previously a slight typo in one argument (`pacakge_data` instead of `package_data`) which is now fixed. I've also added the additional requirements files as optional dependencies which can be installed with `pip` using the extras syntax, for example `pip install s2fft[tests]`. I also integrated the previous content in `pytest.ini` in to the `pyproject.toml` file.


matt-graham:
This PR is a dependency for #204 and we would ideally also merge before resolving #205 as then we can integrate the formatter configuration directly in to the `pyproject.toml` file, so tagging you for a review @CosmoMatt!

I've merged in latest changes from `main` (there was a conflict due to update of version specifier `setup.py` on `main` but we should be able to safely ignore this as version will be automatically derived from latest Git release tag when this goes in) and also removed the generated `s2fft/_version.py` file that was added in 95ea4f591650ef4a30970f059960977f5f8cc390 as this file shouldn't be version controlled (it will be generated on the fly at package installation by `setuptools_scm` based on current latest Git tag).

CosmoMatt:
@matt-graham I had some issues here relating to healpy dropping support for older MacOS versions. I also ran into a version control issue relating to SciPy and healpy, a solution to which was just pinning the SciPy version. This should be fine for now, we'll just need to keep an eye on that in future.

CosmoMatt (approved):
Looks good to me! Only change was pinning the SciPy version to avoid version conflicts.

PR #174, opened 2023-12-06
add explicit doc link to readme

CosmoMatt:
Superficial changes to README. Adds:
- another explicit link to documentation
- explicit link to documentation notebook

This PR can be merged and then the healpix updates can be released in a new version.


jasonmcewen (changes requested):
Looks good.  Only thing is the notebook in the documentation is not executed.  

What is the best policy here?  Ideally it would be good to have notebooks in the repo that have the output cleared (it then much easier to diff).  However, it would be good to have an executed notebook in the documentation.  

I suppose we could get the CI to execute the notebooks before deploying the documentation?  We have done this for the machine learning course.  It's relatively easy to do this using `jupyter nbconvert`.  See example where we have done that [here](https://github.com/astro-informatics/course_mlbd/blob/75ee26df60e2038da886e8c3cfef80668da73b00/.github/workflows/deploy.yml#L100C19-L100C37).

jasonmcewen (approved):

PR #175, opened 2023-12-13
increment version in setup and docs

CosmoMatt:
increments version number to v1.0.1 throughout for new release with HEALPix compile/run time changes.


CosmoMatt:
By-passing main protection as superficial version number changes.

PR #182, opened 2024-02-05
Feature/rotations

CosmoMatt:
@jasonmcewen this PR adds:

### Core additions
- [x] A vectorised JAX implementation of the Risbo Wigner d recursion.
- [x] A rotation function based on this recursion, with a precompute (O(L^3) storage) and iterative (O(L^2)) storage
- [x] Explicitly handle north/South Pole singularities within recursion.

### Housekeeping
- [x] Updates documentation appropriately.
- [x] Adds unit tests for JAX recursion
- [x] Adds unit tests for rotation (both with and without precompute
- [x] Adds gradient tests for rotation (against JAX finite differences)
- [x] Need to add notebook outlining how to perform rotation.


CosmoMatt:
@mousset @PhilippMisofCH  just to keep you in the loop. This PR should add the functionality you've requested, so once @jasonmcewen and I have reviewed and merged the code you should be good to go!

jasonmcewen (commented):
@CosmoMatt All looks good @CosmoMatt !  A couple of quick questions:

- What was the reason for adding special cases for the poles?  Did you see a failure there without this?  I seem to recall I've used the Risbo recursion at the poles previously without any problems.
- I don't get the rationale for the `augmentation.py` filename?  What is this suppose to capture?  What about just `rotate.py` or `rotation.py`?

I've approved so we can merge but can you comment on/address the points above before merging.

jasonmcewen (approved):

jasonmcewen:
Apologies for the delay following up @CosmoMatt , only just saw this.  For Risbo can we just stick to the standard.  It would be good if this implementation to match the tried and tested approach.  So if you could remove the special case for the poles that would be good.  

In terms of file name, augmentation is confusing to me so let's change.  rotations.rotate is indeed slighly overkill but is fine really.  Top level rotate is also fine with me.  Not, let's be sure to call this method rotate_flm tho, since we might implement a pixel space rotation at some point.

I've just seen you had an outstanding point for a notebook to demonstate that.  That is a very good idea.  I actually think we should probably add that before merging so it's done.  Could you quickly add a tutorial notebook on this?

CosmoMatt:
@jasonmcewen the above should all be addressed. Can merge if (when) CI passes and you give it the go ahead :)

jasonmcewen:
Looks good @CosmoMatt , please go ahead and merge at your convenience!

PR #185, opened 2024-02-26
add pytorch support for precompute transform

CosmoMatt:
This PR adds PyTorch support for our precompute transforms, amongst other small adjustments.

### Core Code:
- [x] Add PyTorch forward/inverse spherical harmonic transform
- [x] Add PyTorch forward/inverse Wigner transform
- [x] Add PyTorch resampling and quadrature functions.
- [x] Add PyTorch implementation of HEALPix Fast Fourier transforms.

### Unit Tests
- [x] Roundtrip error for PyTorch spherical harmonic and Wigner transforms.
- [x] Regression error for PyTorch spherical harmonic and Wigner transforms (against base implementations).
- [x] Validation of PyTorch gradients against finite differences 

### Extras
- [x] Notebook demonstrating how to interface with PyTorch frontend. 

### Contributions for next PRs
- [ ] Speed benchmarking of PyTorch implementation against JAX on GPU.
- [ ] Add PyTorch support for recursion based transforms.

### Further comments
HEALPix FFTs require modular indexing into torch tensor, which results in a non-unique tensor of index locations at which to update an existing tensor in place. This doesn't currently work with the expected array[index_array] += new_array behaviour so a workaround was needed. If this changes in future we should update accordingly.


CosmoMatt:
@all-contributors please add @kmulderdas  for bug

.pip_readme.rst
```diff
@@ -27,6 +27,10 @@ for adjoint transformations where needed, and comes with different
 optimisations (precompute or not) that one may select depending on
 available resources and desired angular resolution $L$.
 
+As of version 1.0.2 `S2FFT` also provides PyTorch implementations of underlying 
+precompute transforms. In future releases this support will be extended to our 
+recursion based algorithms.
```
jasonmcewen, line 32: The terminology we use in the paper in on-the-fly algorithms.  (Since the pre-compute is essentially also recursion based, we just precompute the recursion.)

README.md
```diff
@@ -22,6 +22,10 @@ for adjoint transformations where needed, and comes with different
 optimisations (precompute or not) that one may select depending on
 available resources and desired angular resolution $L$.
 
+As of version 1.0.2 `S2FFT` also provides PyTorch implementations of underlying 
+precompute transforms. In future releases this support will be extended to our 
+recursion based algorithms.
+
```
jasonmcewen, line 28: Same comment as above.

docs/index.rst
```diff
@@ -10,6 +10,10 @@ transforms (for both real and complex signals), with support for adjoint transfo
 where needed, and comes with different optimisations (precompute or not) that one
 may select depending on available resources and desired angular resolution :math:`L`.
 
+As of version 1.0.2 ``S2FFT`` also provides PyTorch implementations of underlying 
+precompute transforms. In future releases this support will be extended to our 
+recursion based algorithms.
```
jasonmcewen, line 15: Same comment as above.

s2fft/utils/quadrature_torch.py
```diff
@@ -0,0 +1,256 @@
...
+        ValueError: Invalid sampling scheme.
+
+    Returns:
+        torch.tensor: Quadrature weights *to use in transform* for sampling scheme for
+        each :math:`\theta` (weights are identical as :math:`\phi` varies for given
+        :math:`\theta`).
+    """
+
+    if sampling.lower() == "mwss":
+        return quad_weights_mwss_theta_only(2 * L) * 2 * torch.pi / (2 * L)
```
jasonmcewen, line 34: Can you add some extra tests catch these and following cases (see codecov comments).

jasonmcewen (changes requested):
Looks great @CosmoMatt !  Just some minor comments on the documentation.  And could you add a few more tests to catch version cases.  On the code diff in github, a number of lines are flagged by codecov as not being tested.  Would be great to quickly add some tests for these cases so we keep the test coverage as high as possible.  Once these very minor points are addressed then I think we're good to merge and do a new release.

.pip_readme.rst
CosmoMatt: updated.

README.md
CosmoMatt: updated.

docs/index.rst
CosmoMatt: updated.

CosmoMatt:
@jasonmcewen difficult to test the HEALPix quadrature authentically as its not exact and shouldn't pass an f00 test. I did however extend the testing to cover DH. Overall should be at around 94% which is fine I think.

jasonmcewen (approved):

PR #204, opened 2024-05-22
Cufft primitive

ASKabalan:
Creating a pure cuda healpix fft to avoid the long compile time when using pure JAX


lib/include/kernel_nanobind_helpers.h
matt-graham, line 1: Again we need to be careful here about including a file from another project under a different license (Apache 2.0). For example there is a term in the Apache-2.0 license 

```
      (b) You must cause any modified files to carry prominent notices
          stating that You changed the files; and
```

There does appear to be some minor modifications here (for example name of namespace, adding/removing includes) but we don't have such a prominent notice.

Ideally we'd avoid having a copy of this file at all by just including the existing `jaxlib` file but I'm not sure how feasible that is to do robustly. Otherwise I think we need to add some sort of notice about the modiciations made.

s2fft/utils/jax_pritimive.py
```diff
@@ -0,0 +1,136 @@
+from abc import ABCMeta, abstractmethod
+from dataclasses import dataclass
+from functools import partial
+from typing import Tuple
+
+import jax.numpy as jnp
+from jax import jit, lax
+from jax._src.api import ShapeDtypeStruct
+from jax._src.core import ShapedArray
+from jax._src.typing import Array, ArrayLike
```
matt-graham, line 10: We should be primarily  importing from JAX's [public API ](https://jax.readthedocs.io/en/latest/api_compatibility.html) not from the `jax._src` subpackage which is not considered part of the public API and can be changed without warning. While we may need to rely on some modules not in the public API, these types all seem to be available in the public API under different modules so should be imported from there.

s2fft/utils/jax_pritimive.py
```diff
@@ -0,0 +1,136 @@
...
+import operator
+import os
+import warnings
+
+import numpy as np
+import jax.numpy as jnp
+from jax.lib import xla_client
+from jax import core, dtypes
+from jax.interpreters import xla, mlir
+from jax.experimental.custom_partitioning import custom_partitioning
```
matt-graham, line 28: This is a repeat of an import stament above.

s2fft/utils/healpix_ffts.py
```diff
@@ -5,10 +5,24 @@
...
+from s2fft.utils.jax_pritimive import BasePrimitive, register_primitive
+import jaxlib.mlir.ir as ir
+from s2fft_lib import _s2fft
+from jaxlib.hlo_helpers import custom_call
+from jax._src.lib.mlir.dialects import hlo
+from jax._src.interpreters import mlir
+from jax.core import Primitive, ShapedArray
+from jax.interpreters import ad, xla
+from jax.sharding import Mesh, NamedSharding
+from jax.sharding import PartitionSpec as P
```
matt-graham, line 17: This appears to be unused - also I would avoid aliasing classes to single character names.

s2fft/utils/healpix_ffts.py
```diff
@@ -5,10 +5,24 @@
...
 from s2fft.sampling import s2_samples as samples
 from functools import partial
+from s2fft.utils.jax_pritimive import BasePrimitive, register_primitive
+import jaxlib.mlir.ir as ir
+from s2fft_lib import _s2fft
+from jaxlib.hlo_helpers import custom_call
+from jax._src.lib.mlir.dialects import hlo
+from jax._src.interpreters import mlir
+from jax.core import Primitive, ShapedArray
+from jax.interpreters import ad, xla
```
matt-graham, line 15: These imports appear to be unused

s2fft/utils/healpix_ffts.py
```diff
@@ -5,10 +5,24 @@
...
+from jax.sharding import Mesh, NamedSharding
```
matt-graham, line 16: `NamedSharding` appears to be unused.

s2fft/utils/healpix_ffts.py
```diff
@@ -5,10 +5,24 @@
...
+from jax.sharding import PartitionSpec as P
+from typing import Tuple
+from jax._src.api import ShapeDtypeStruct
+from math import prod
```
matt-graham, line 20: This appears to be unused.

s2fft/utils/healpix_ffts.py
```diff
@@ -5,10 +5,24 @@
...
+from s2fft_lib import _s2fft
+from jaxlib.hlo_helpers import custom_call
+from jax._src.lib.mlir.dialects import hlo
+from jax._src.interpreters import mlir
+from jax.core import Primitive, ShapedArray
+from jax.interpreters import ad, xla
+from jax.sharding import Mesh, NamedSharding
+from jax.sharding import PartitionSpec as P
+from typing import Tuple
+from jax._src.api import ShapeDtypeStruct
```
matt-graham, line 19: We should avoid importing anything from `jax._src` as its not part of public API.

matt-graham (commented):
Thanks @ASKabalan for your efforts on this - you've done lots of excellent work here 🎉. I've added some initial high-level comments which are mainly around how this packaged and structured.

The most significant issue at the moment is I think the currently implementation precludes building / installing the package on systems without CUDA which I think we don't want to do - I get a CMake error
```
      CMake Error at /usr/share/cmake-3.22/Modules/CMakeDetermineCUDACompiler.cmake:179 (message):
        Failed to find nvcc.

        Compiler requires the CUDA toolkit.  Please set the CUDAToolkit_ROOT
        variable.
```
when trying to `pip install` the packages with edits here locally in an environment without CUDA.

You've also included some reformatting changes to the `s2fft/utils/healpix_ffts.py` file which make it difficult to see the more substantive changes you made. Ideally any automatic reformatting should be done in a separate PR to keep the diffs easier to read here and after we've had some discussion about the relevant style guide to follow and formatter to use, so would it be possible for you to revert the formatting changes in that file to make this easier to review.

.gitmodules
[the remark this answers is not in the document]
ASKabalan: I think that this is a choice.
JAX uses this kind of mecanism with bazel.
We used submodules in jaxDecomp
I think I will leave the choice to you.
If you want a pip package, I will do a pip package

lib/include/argparse.hpp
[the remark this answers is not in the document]
ASKabalan: I think that I used this header for testing purposes.
Before linking with python.
Do you think that it is necessary to have a standalone testing executable?
again your choice

CMakeLists.txt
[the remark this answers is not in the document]
ASKabalan: Yes this is a good thing to do.
However I  am aiming to make it even easier .
Two things to keep in mind

- For GPU, you need nvcc to compile. I can use the pip installable nvcc to compile and install the nvidia wheels (or use those installed by JAX to compile)
- Build the wheels ourselves and somehow push them to pypi (for linux only for example) just like JAX

The second I don't know how to do yet, but I am gonna do it for jaxDecomp.
Basically it means that the users installs pre-built s2fft.so library

s2fft/utils/healpix_ffts.py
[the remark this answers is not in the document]
ASKabalan: should I make a formatting PR?.
Using a style.yapf makes it easier for all of us to share the same formatting

.style.yapf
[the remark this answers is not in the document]
ASKabalan: I don't mind using any formatter.
As long as we use the same one
Tell me what you prefer and I will make a PR

s2fft/utils/healpix_ffts.py
ASKabalan: Not possible for this exact case.
I would look for an alternative
But when writing primitives, we have to dig deep in _src

matt-graham: > Not possible for this exact case.
> I would look for an alternative

`ShapeDtypeStruct` is [available under the top-level `jax` namespace](https://jax.readthedocs.io/en/latest/_autosummary/jax.ShapeDtypeStruct.html).

> But when writing primitives, we have to dig deep in _src

I don't think this is true as for example [the `kepler_jax.py` file](https://github.com/dfm/extending-jax/blob/main/src/kepler_jax/kepler_jax.py) in the `extending-jax` tutorial repository does not have any imports from under the `jax._src` subpackage.

ASKabalan: I am gonna make a second pass on the file and clean up
But I started from my own style of primitives see [here](https://github.com/DifferentiableUniverseInitiative/jaxDecomp/blob/main/jaxdecomp/_src/transpose.py)
Which itself is inspired from nvidia style [here](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/jax/cpp_extensions/normalization.py#L40)
Which is more up to date and also has the nessecary language to define multi-gpu primitives that DFM's example does not do.
Some of the imports are used as argument type hints .. for [example in my code](https://github.com/DifferentiableUniverseInitiative/jaxDecomp/blob/main/jaxdecomp/_src/transpose.py#L153)
Whereas in NVIDIA's [code](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/jax/cpp_extensions/normalization.py#L185)  it is not used (so no need for import)

Nevertheless, I will make a pass and remove `_src` imports

s2fft/utils/healpix_ffts.py
ASKabalan: For multi-gpu 
I can remove it if you want

s2fft/utils/healpix_ffts.py
ASKabalan: Multi-gpu + type hinting stuff.
Again, I can remove if necessary
Except for prod ... that was just a copy paste from my own work in jaxDecomp (will be removed)

s2fft/utils/healpix_ffts.py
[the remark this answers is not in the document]
matt-graham: I've opened an issue #205 about formatting / code style convention.

ASKabalan:
I have implemented what you asked

- use nanobind as pip package and not submodule
- remove _src imports when possible
- to be able to compile without cuda (but show error message when the user tries to call cuda functions)

I think that this PR can be merged (if it is all ok for you)
Things we need to take care of 

- [ ] removing dependency on pyssht if it is failing https://github.com/astro-informatics/s2fft/issues/206
- [ ] formatting https://github.com/astro-informatics/s2fft/issues/205

ASKabalan:
> Hi @ASKabalan. Thanks for the updates to the pull-request, we're hoping to try to get this merged in as soon as possible.
> 
> At the moment several (previously passing) tests in `tests/test_healpix_ffts.py` are failing because of a mismatch between the output of `healpix_ifft_jax` and `healpix_ifft_numpy`. The whitespace changes in `s2fft/utils/healpix_ffts.py` make it difficult to see where the actual substantive code changes in this file are which might have caused these tests to start failing. Could I ask that you revert the changes which are not connected to this PR to help us identify where this issue is arising?
> 
> We also need to have the tests which require a CUDA GPU to be skipped when running on non-GPU systems (such as the GitHub Action test runners). I can update the tests to do this if you are okay for me to commit to this PR branch directly.

I suggest doing a seperate formatting PR from the main branch
Format using a pre-commit formatter
Then I do the same and merge this way the comparaison is easier


I changed one thing in the JAX ifft function.
This https://github.com/astro-informatics/s2fft/blob/main/s2fft/utils/healpix_ffts.py#L478
to https://github.com/astro-informatics/s2fft/blob/cufft-primitive/s2fft/utils/healpix_ffts.py#L470

Using the actual IFFT instead of the FFT of the complex conjugate
We discussed this in the slack channel

LICENSE.txt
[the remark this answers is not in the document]
ASKabalan: I did it because It was blocking the installation for some reason at some point
I will revert

matt-graham:
> I suggest doing a seperate formatting PR from the main branch Format using a pre-commit formatter Then I do the same and merge this way the comparaison is easier

We can so a separate formatting PR and merge this in first, but this is likely to result in quite messy merge conflicts as if we are reformatting the whole codebase then we will be touching a lot of lines in the same files that are being updated in this PR. If you prefer to go this way we can do so, but I would say reverting the white space changes here would be a lot simpler.

> I changed one thing in the JAX ifft function. This https://github.com/astro-informatics/s2fft/blob/main/s2fft/utils/healpix_ffts.py#L478 to https://github.com/astro-informatics/s2fft/blob/cufft-primitive/s2fft/utils/healpix_ffts.py#L470
> 
> Using the actual IFFT instead of the FFT of the complex conjugate We discussed this in the slack channel

My understanding from the conversation in Slack was that the change shouldn't alter the numerical output, it was just meant to be removing a previous workaround for problems in 64-bit precision with the underlying JAX IFFT implementation? Either way as the test is failing, either the test or implementation need updating.

Looking at the change from

```Python
jnp.conj(jnp.fft.fft(jnp.fft.ifftshift(jnp.conj(fm_chunks), axes=-1), norm="backward")        
```

to

```Python
jnp.fft.ifft(jnp.fft.ifftshift(fm_chunks, axes=-1), norm="backward")
```

I think the problem maybe a difference in normalization factors as [this note](https://adamsiembida.com/how-to-compute-the-ifft-using-only-the-forward-fft/) suggests the equivalence is

$$
  \mathsf{IFFT}(X) = \frac{1}{N} \mathsf{conj}(\mathsf{FFT}(\mathsf{conj}(\mathsf{FFT}(X)))
$$

on the basis of a $\frac{1}{N}$ normalization factor being included in `jnp.fft.ifft` and no normalization in `jnp.fft.fft`, which from the [NumPy documentation](https://numpy.org/doc/stable/reference/routines.fft.html#normalization) I think should correspond to the default behaviour when `norm` arguments are not specified or when `norm = "backward"` (and I believe JAX should follow NumPy behaviour).

To retain the previous behaviour I think we need to account for the $\frac{1}{N}$ factor, either explicitly or by setting the `norm` argument to `ifft` accordingly. Looking at your changes, you appear to have `norm = "backward"` in `healpix_ifft_jax` while `healpix_ifft_numpy` uses `norm = "forward"` so that suggests you need to update to follow the NumPy implementation here, or if we decide on a different normalization convention we will need to update all of the IFFT function to use the same convention, though as this would be a breaking change I think we should avoid if possible.

ASKabalan:
@matt-graham You were right.
I fixed the issue all tests are good now
I also added a condition on the test to run only if GPU is available.

pyproject.toml
matt-graham, line 1: The updates to this file mean the package dependencies and optional dependencies are not currently being specified. Previously these where set dynamically using the `tool.setuptools.dynamic`, with the version also being set dynamically by `setuptools_scm`. It appears `scikit-build-core` does not support dynamics (optional) dependencies, so we will instead need to include them directly in this file and remove the corresponding requirements files (to avoid duplication). [Using `setuptools_scm` to set the version dynamically is compatible with `scikit-build-core`](https://scikit-build-core.readthedocs.io/en/latest/configuration.html#dynamic-metadata) though so we should retain that. Changes required to `pyproject.toml` are 

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 0f903be..60ea207 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,15 +25,58 @@ classifiers = [
     "Intended Audience :: Science/Research",
 ]
 description = "Differentiable and accelerated spherical transforms with JAX"
+dependencies = [
+    "numpy>=1.20,<2",
+    "colorlog",
+    "pyyaml",
+    "jax>=0.3.13",
+    "jaxlib",
+    "torch",
+    "pyssht",
+    "healpy",
+    "ducc0",
+]
+dynamic = [
+    "version"
+]
 keywords = [
 ]
 name = "s2fft"
-version = "0.1.0"
 readme = ".pip_readme.rst"
 requires-python = ">=3.8"
 license.file = "LICENSE.txt"
 urls.homepage = "https://github.com/astro-informatics/s2fft"

+[project.optional-dependencies]
+docs = [
+    "sphinx>=5.0.0",
+    "nbsphinx-link>=1.3.0",
+    "pydata-sphinx-theme>=0.12.0",
+    "sphinx_toolbox>=2.15.0",
+    "sphinx-tabs>=3.2.0",
+    "sphinxcontrib-bibtex>=2.4.1",
+    "sphinx-git>=11.0.0",
+    "sphinxcontrib-texfigure>=0.1.3",
+    "sphinx-copybutton",
+    "sphinxemoji",
+    "sphinx_rtd_theme",
+    "ipython>=7.16.1",
+    "jupyter>=1.0.0",
+]
+plotting = [
+    "pyvista",
+    "trame",
+    "ipykernel",
+    "ipywidgets",
+]
+tests = [
+    "pytest",
+    "pytest-cov",
+    "codecov",
+    "so3",
+]
+
+
 [tool.scikit-build]
 # Protect the configuration against future changes in scikit-build-core
 minimum-version = "0.4"
@@ -44,6 +87,9 @@ build-dir = "build/{wheel_tag}"
 # Build stable ABI wheels for CPython 3.12+
 wheel.py-api = "cp312"

+metadata.version.provider = "scikit_build_core.metadata.setuptools_scm"
+sdist.include = ["s2fft/_version.py"]
+
 [tool.pytest.ini_options]
 addopts = "--color=yes -v"
 testpaths = [
@@ -58,6 +104,7 @@ filterwarnings = [

 [tool.setuptools]
 packages = ["s2fft"]
+write_to = "s2fft/_version.py"

 [tool.setuptools.dynamic]
 dependencies = {file = ["requirements/requirements-core.txt"]}
```

ASKabalan:
> ## [Codecov](https://app.codecov.io/gh/astro-informatics/s2fft/pull/204?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=astro-informatics) Report
> 
> Attention: Patch coverage is `59.84848%` with `53 lines` in your changes missing coverage. Please review.
> 
> > Project coverage is 91.76%. Comparing base [(`2c3d9e5`)](https://app.codecov.io/gh/astro-informatics/s2fft/commit/2c3d9e5af52940846b265d8912fc7ab0296f6891?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=astro-informatics) to head [(`c552ce5`)](https://app.codecov.io/gh/astro-informatics/s2fft/commit/c552ce5718aea4a3377813c8214ba66bc7dd05d0?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=astro-informatics).
> 
> [Files with missing lines](https://app.codecov.io/gh/astro-informatics/s2fft/pull/204?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=astro-informatics) 	Patch % 	Lines
> [s2fft/utils/healpix_ffts.py](https://app.codecov.io/gh/astro-informatics/s2fft/pull/204?src=pr&el=tree&filepath=s2fft%2Futils%2Fhealpix_ffts.py&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=astro-informatics#diff-czJmZnQvdXRpbHMvaGVhbHBpeF9mZnRzLnB5) 	38.35% 	[45 Missing ⚠️ ](https://app.codecov.io/gh/astro-informatics/s2fft/pull/204?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=astro-informatics)
> [s2fft/utils/jax_pritimive.py](https://app.codecov.io/gh/astro-informatics/s2fft/pull/204?src=pr&el=tree&filepath=s2fft%2Futils%2Fjax_pritimive.py&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=astro-informatics#diff-czJmZnQvdXRpbHMvamF4X3ByaXRpbWl2ZS5weQ==) 	86.44% 	[8 Missing ⚠️ ](https://app.codecov.io/gh/astro-informatics/s2fft/pull/204?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=astro-informatics)
> Additional details and impacted files
> 
> 		
> 		
> 
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/astro-informatics/s2fft/pull/204?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=astro-informatics). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=astro-informatics).

The tests are covered.
The fact that there is no GPU does not trigger codecov?

tests/test_healpix_ffts.py
```diff
@@ -1,48 +1,86 @@
 import numpy as np
+from numpy.testing import assert_allclose
 import healpy as hp
 import pytest
 import jax
+if jax.__version__ < "0.4.32":
```
matt-graham, line 6: String comparison on version specifiers can give incorrect results - [for example](https://stackoverflow.com/q/11887762)

```
>>> "2.3.1" > "10.1.1"
True
```

To be more robust here we would need to something like:

```Python
from packaging.version import Version as _Version

...

if _Version(jax.__version__) < _Version("0.4.32"):
```

and add `packaging` to project dependencies.

matt-graham:
> The tests are covered. The fact that there is no GPU does not trigger codecov?

The codecov results are based on the coverage reported from the GitHub Actions workflow and so won't include the tests you added which are now only run when a GPU is available. I think we can just ignore the codecov checks here or [set the corresponding functions to be ignored in the coverage reports](https://coverage.readthedocs.io/en/latest/excluding.html).

jasonmcewen:
> > I changed one thing in the JAX ifft function. This https://github.com/astro-informatics/s2fft/blob/main/s2fft/utils/healpix_ffts.py#L478 to https://github.com/astro-informatics/s2fft/blob/cufft-primitive/s2fft/utils/healpix_ffts.py#L470
> > Using the actual IFFT instead of the FFT of the complex conjugate We discussed this in the slack channel
> 
> My understanding from the conversation in Slack was that the change shouldn't alter the numerical output, it was just meant to be removing a previous workaround for problems in 64-bit precision with the underlying JAX IFFT implementation? Either way as the test is failing, either the test or implementation need updating.
> 
> Looking at the change from
> 
> ```python
> jnp.conj(jnp.fft.fft(jnp.fft.ifftshift(jnp.conj(fm_chunks), axes=-1), norm="backward")        
> ```
> 
> to
> 
> ```python
> jnp.fft.ifft(jnp.fft.ifftshift(fm_chunks, axes=-1), norm="backward")
> ```
> 
> I think the problem maybe a difference in normalization factors as [this note](https://adamsiembida.com/how-to-compute-the-ifft-using-only-the-forward-fft/) suggests the equivalence is
> 
> IFFT ( X ) = 1 N conj ( FFT ( conj ( FFT ( X ) ) )
> 
> on the basis of a 1 N normalization factor being included in `jnp.fft.ifft` and no normalization in `jnp.fft.fft`, which from the [NumPy documentation](https://numpy.org/doc/stable/reference/routines.fft.html#normalization) I think should correspond to the default behaviour when `norm` arguments are not specified or when `norm = "backward"` (and I believe JAX should follow NumPy behaviour).
> 
> To retain the previous behaviour I think we need to account for the 1 N factor, either explicitly or by setting the `norm` argument to `ifft` accordingly. Looking at your changes, you appear to have `norm = "backward"` in `healpix_ifft_jax` while `healpix_ifft_numpy` uses `norm = "forward"` so that suggests you need to update to follow the NumPy implementation here, or if we decide on a different normalization convention we will need to update all of the IFFT function to use the same convention, though as this would be a breaking change I think we should avoid if possible.
> 
> EDIT: Tagging @CosmoMatt for comment here as he is probably best placed to advise on what the normalization conventions we should be following are!

Just had a chance to catch up on these discussions.  It seems like this has all been resolved now but for further background, @CosmoMatt found that `jnp.fft.ifft` wasn't fully supported in JAX(!) back when we first implemented this so as a workaround he implemented the IFFT via `jnp.fft.fft`.  It seems the JAX bug was resolved in more recent versions of JAX and so the code could be tidied up to use `jnp.fft.ifft` directly.  (@CosmoMatt was this applied anywhere else that we should also update now?)

The line itself shouldn't apply any normalisation (that is handled elsewhere), so using an `jnp.fft.fft` with `norm="backward"` achieves that.  Since we now switch to `jnp.fft.ifft` we should switch to `norm="forward"`, which ensures no normalisation is applied in that line of code.  

In general, I think we've tried to follow the convention where `jnp.fft.fft` and `jnp.fft.ifft` are performed without any normalisation and we instead apply the appropriate normalisation explicitly ourselves since often there are additional normalisation factors to be considered.  That corresponds to setting `norm="backward"` when calling `jnp.fft.fft` and setting `norm="forward"` when calling `jnp.fft.ifft` (although I couldn't say for sure this approach is applied absolutely everywhere in the code but that is least the generally approach we've typically been adopting).

lib/include/argparse.hpp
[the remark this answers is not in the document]
matt-graham: Having the option to build a standalone test executable is I think nice given you've already done the effort of getting this working. I would still say better to fetch a tagged version in `CMakeLists.txt` than including source directly though.

pyproject.toml
```diff
@@ -0,0 +1,117 @@
...
+[tool.setuptools]
+packages = ["s2fft"]
+write_to = "s2fft/_version.py"
+
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements/requirements-core.txt"]}
+optional-dependencies.docs = { file = ["requirements/requirements-docs.txt"] }
+optional-dependencies.plotting = { file = ["requirements/requirements-plotting.txt"] }
+optional-dependencies.tests = { file = ["requirements/requirements-tests.txt"] }
+
```
matt-graham, line 110:
```suggestion
```
This section is now redundant with dependencies included directly in file above. The corresponding files in `requirements` directory can also be removed.

requirements/requirements-core.txt
```diff
@@ -8,3 +8,4 @@ torch # Remove when subpackage functionality is fixed.
 pyssht
 healpy
 ducc0
+nanobind >=2.0.0
```
matt-graham, line 11: We should be able to remove this file now anyway, but I don't think we have a runtime dependency on `nanobind` anyway?

s2fft/_version.py
matt-graham, line 1: This file was mistakenly added in a previous commit and was removed in #173 - may need to resolve conflict by just deleting file when merging in changes from `main`.

s2fft/transforms/spherical.py
```diff
@@ -537,7 +542,7 @@ def forward_numpy(
     return flm * (-1) ** spin
 
 
-@partial(jit, static_argnums=(1, 3, 4, 5, 7, 8))
+@partial(jit, static_argnums=(1, 3, 4, 5, 7, 8,9))
```
matt-graham, line 545:
```suggestion
@partial(jit, static_argnums=(1, 3, 4, 5, 7, 8, 9))
```

requirements/requirements-core.txt
ASKabalan: No it is only build time
I will remove it

s2fft/transforms/spherical.py
```diff
@@ -582,6 +588,8 @@ def forward_jax(
         L_lower (int, optional): Harmonic lower-bound. Transform will only be computed
             for :math:`\texttt{L_lower} \leq \ell < \texttt{L}`. Defaults to 0.
 
+        use_cuda (bool, optional): Whether to use the CUDA backend. Defaults to False.
+
```
matt-graham, line 592: I think it would be worth expanding on what this flags does a bit more in this docstring and/or using a more informative name. My understanding is this only has an effect when `sampling == "healpix"` and even in that case when `use_cuda = False` the computation will still use a (CUDA-compatible) GPU if available, but will use inbuilt JAX primitives rather than the custom primitive being added in this PR. I think something like `use_healpix_custom_primitive` would be more a explicit name and changing docstring to something like

```
        use_healpix_custom_primitive (bool, optional): Whether to use a custom CUDA primitive for
            computing HEALPix fast fourier transform when `sampling = "healpix"` and running on a
            CUDA compatible GPU device. Using a custom primitive reduces long compilation times
            when JIT compiling. Defaults to `False`.
```

matt-graham:
Hi @ASKabalan. Thanks for your updates to the PR. There are still some previous review comments which I think are unresolved I've catalogued below and I've added some further comments from another review

- [ ] Commented out code in `s2fft/utils/healpix_ffts.py` [(link to comment)](https://github.com/astro-informatics/s2fft/pull/204/files#r1638338763)
- [ ] Fixing typo in name of `s2fft/utils/jax_pritimive.py` file [(link to comment)](https://github.com/astro-informatics/s2fft/pull/204/files#r1638277123)
- [ ] Fixing indentation in docstrings in `s2fft/utils/jax_pritimive.py` file [(link to comment)](https://github.com/astro-informatics/s2fft/pull/204/files#r1638279292)
- [ ] Fixing broken URL in comment in `s2fft/utils/jax_pritimive.py` [(link to comment)](https://github.com/astro-informatics/s2fft/pull/204/files#r1638280737)
- [ ] Resolving use of external code by adjusting copyright statement in license or removing use of `BasePrimitive` abstract base class [(link to comment)](https://github.com/astro-informatics/s2fft/pull/204/files#r1638338763)
- [ ] Resolving licensing for  external code in `lib/include/kernel_helpers.h` [(link to comment)](https://github.com/astro-informatics/s2fft/pull/204#discussion_r1638357861)
- [ ] Adjusting license in header for `lib/include/kernel_nanobind_helpers.h` to list modifications made [(link to comment)](https://github.com/astro-informatics/s2fft/pull/204#discussion_r1638376027)
- [ ] Fetching `argparse` library in `CMakesLists.txt` rather than including in repository [(link to comment)](https://github.com/astro-informatics/s2fft/pull/204#discussion_r1638143845)
- [ ] Remove `tools.setuptools.dynamic` section from `pyproject.toml` and remove redundant `requirements` files [(link to commment)](https://github.com/astro-informatics/s2fft/pull/204/files#r1782453077)
- [ ] Updating docstring and argument name for `use_cuda` argument [(link to comment)](https://github.com/astro-informatics/s2fft/pull/204/files#r1782473001)
- [ ] Revert renaming of `LICENCE.txt` to `LICENSE.txt` [(link to comment)](https://github.com/astro-informatics/s2fft/pull/204#discussion_r1771057518)

If you prefer I am happy to deal with addressing these.

s2fft/transforms/spherical.py
ASKabalan: Didn't want to format and ruin the entire file :/

pyproject.toml
```diff
@@ -52,12 +103,7 @@ filterwarnings = [
 
 [tool.setuptools]
 packages = ["s2fft"]
-
-[tool.setuptools.dynamic]
-dependencies = {file = ["requirements/requirements-core.txt"]}
-optional-dependencies.docs = { file = ["requirements/requirements-docs.txt"] }
-optional-dependencies.plotting = { file = ["requirements/requirements-plotting.txt"] }
-optional-dependencies.tests = { file = ["requirements/requirements-tests.txt"] }
+write_to = "s2fft/_version.py"
```
matt-graham, line 106:
```suggestion
```
This is I think should belong under `tool.setuptools_scm` and it appears there is already a entry for this below

pyproject.toml
```diff
@@ -24,6 +26,17 @@ classifiers = [
...
+    "jax>=0.3.13",
+    "jaxlib",
+    "torch",
+    "pyssht",
+    "healpy",
+    "ducc0",
+]
 dynamic = [
     "version",
     "dependencies",
```
matt-graham, line 42:
```suggestion
```
I think this is what is causing tests workflow to fail.

s2fft/utils/jax_pritimive.py
```diff
@@ -0,0 +1,136 @@
...
+from jax.interpreters.mlir import ir, dtype_to_ir_type
+from jax._src.interpreters import batching
+from jax._src import dispatch
+
+# Inspired by https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/jax/cpp_extensions.py
+# and by the tweaked version in https://github.com/DifferentiableUniverseInitiative/jaxDecomp/blob/main/jaxdecomp/_src/spmd_ops.py
```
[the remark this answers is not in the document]
matt-graham: The issue here is that we can't include a derivative version of the code from `NVIDIA/TransformerEngine` repository which is under [the Apache 2.0 License](https://github.com/NVIDIA/TransformerEngine/blob/release_v1.7/LICENSE) unless we meet the conditions of redistribution under the terms of that license, which in particular require us to

>       (a) You must give any other recipients of the Work or
>           Derivative Works a copy of this License; and
> 
>       (b) You must cause any modified files to carry prominent notices
>           stating that You changed the files; and
> 
>       (c) You must retain, in the Source form of any Derivative Works
>           that You distribute, all copyright, patent, trademark, and
>           attribution notices from the Source form of the Work,
>           excluding those notices that do not pertain to any part of
>           the Derivative Works; and

So here we either need to 

  1. include a copy of the original license, a notice stating we modified the file and an attribution to the original copyright  holder (NVIDIA),
  2. or remove this code.

I would favour the latter, because I don't think there is much advantage to having an abstract `BasePrimitive` class here. We could get similar code modularity just with a function something like

```Python
def register_primitive(
    name: str,
    multiple_results: bool,
    abstract_evaluation: Callable,
    lowering_per_platform: dict[None | str, Callable],
    batcher: None | Callable = None,
    jacobian_vector_product: None | Callable = None,
    transpose: None | Callable = None,
):
    primitive = core.Primitive(name)
    primitive.multiple_results = multiple_results
    primitive.def_impl(partial(xla.apply_primitive, primitive))
    primitive.def_abstract_eval(abstract_evaluation)
    for platform, lowering in lowering_per_platform.items():
        mlir.register_lowering(primitive, lowering, platform=platform)
    if batcher is not None:
        batching.primitive_batchers[primitive] = batcher
    if jacobian_vector_product is not None:
        ad.primitive_jvps[primitive] = jacobian_vector_product
    if transpose is not None:
        ad.primitive_transposes[primitive] = transpose
    return primitive
```

ASKabalan:
@matt-graham  It is ok for me .. go for merge?

matt-graham:
> @matt-graham It is ok for me .. go for merge?

The failing tests are as I forgot we're targetting Python 3.9 and above and I used some Python 3.10 syntax in the type hints. I will fix this now. Then hopefully if tests then pass from my side I'm good to merge, but also tagging @jasonmcewen and @CosmoMatt in case they want to do any final reviews.

ASKabalan:
@matt-graham Also I think that the test should be marked to not be included in the codecov workflow
Should i do it?

matt-graham:
@all-contributors please add @ASKabalan for code, review, test

matt-graham (approved):

PR #210, opened 2024-09-16
Update test matrix and add scheduled tests

### LINKED ISSUE #208: Add weekly tests to catch breaks due to updates in dependencies

matt-graham:
Updates test matrix to run tests across all supported Python versions (currently Python 3.8 &ndash; 3.12) ~~and operating systems~~. Also adds weekly scheduled test runs to resolve #208.

Given this makes the test matrix quite large ~~(15 combinations of OS and Python version)~~, and Python 3.8 and Python 3.9 are both out of the recommended support window of both [SPEC0 ](https://scientific-python.org/specs/spec-0000/#support-window) and [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table), I would also recommend we consider dropping Python 3.8 and Python 3.9 support (in a separate PR).

EDIT: Appears that there are not Windows wheels available for `healpy` / `pyssht` and we're getting a segfault somewhere when using MacOS / Python 3.9+ so reverting to only testing on Ubuntu + (MacOS, Python 3.8) for now.


matt-graham:
I've merged in `main` to resolve conflicts from some files also edited in #233 (mainly some manual edits to make docstrings raw here to avoid issues on newer Python versions that Ruff also flagged and fixed). I've also added a `path` filter to the pull-request trigger here to only run tests workflow when we are updating files which could lead to changes in test behaviour, which should prevent needing to run tests when for example just updating Markdown files or metadata files like `all-contributorsrc`.

CosmoMatt (approved):
Weird that we're running into issues with the actions MacOS workers, but we can address this later. In any case, great to set up testing across a variety of architectures, and importantly the nightly tests!

PR #213, opened 2024-09-16
Use codecov-action in place of manual upload

### LINKED ISSUE #211: `codecov` API token exposed in workflow file
Currently the Codecov API token is included directly in the workflow file:

https://github.com/astro-informatics/s2fft/blob/2c3d9e5af52940846b265d8912fc7ab0296f6891/.github/workflows/tests.yml#L38

In general tokens like this should be stored as a GitHub Actions [secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions) (see for example [this guide in Codecov documentation](https://docs.codecov.com/docs/adding-the-codecov-token](https://docs.codecov.com/docs/adding-the-codecov-token#github-actions))). Ideally we should also use the [`codecov-action` GitHub Actions action](https://github.com/codecov/codecov-action) to upload.

matt-graham:
Would resolve #211 

Will require a repository secret `CODECOV_TOKEN ` to be added with a [Codecov repository upload token](https://docs.codecov.com/docs/frequently-asked-questions#where-is-the-repository-upload-token-found) set to work.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,5 +32,9 @@ jobs:
     
     - name: Run tests
       run: |
-        pytest --cov-report term --cov=s2fft --cov-config=.coveragerc 
-        codecov --token 298dc7ee-bb9f-4221-b31f-3576cc6cb702
+        pytest --cov-report=xml --cov=s2fft --cov-config=.coveragerc
+  
+    - name: Upload coverage reports to Codecov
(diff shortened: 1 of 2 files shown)


CosmoMatt (approved):
Looks fine to me, good to get this hidden in secrets as it otherwise is technically a security risk.

Issue #209, opened 2024-08-14
Inconsistent results for pure JAX implementation of inverse Wigner transformation


ElisR:
Hi! Firstly, thanks a lot for developing this package! I'm not in astro, but the package has been just what I needed in a project I'm working on.

The issue I'm having is related to the JAX implementation of the inverse Wigner transformation `s2fft.wigner.inverse_jax`. I was lucky to stumble upon the SSHT version of this function that allowed me to implement what I needed, albeit on the CPU.

Below, I have added a minimum (non-)working example to illustrate the kind of use case I have.
Apologies that it is still slightly verbose.
I haven't delved deep into the library code to try and debug, but from the scales it looks like some catastrophic floating point errors are happening (beyond aliasing artefacts, which is what I originally assumed).

```python
"""Minimum example showing `s2fft`'s inverse Wigner transformation not working reliably."""

from collections.abc import Callable
from typing import TypeAlias

import jax
import jax.numpy as jnp
import matplotlib.pyplot as plt
from jax.scipy.spatial import transform

# Allowing 64-bit float before importing s2fft
jax.config.update("jax_enable_x64", True)
import s2fft  # noqa: E402  pylint: disable=C0413
from s2fft import partial, sampling  # noqa: E402  pylint: disable=C0413

ScalarFunction: TypeAlias = Callable[[jax.Array], jax.Array]


S2FFT_ANGLE_SAMPLER: str = "dh"
S2FFT_REALITY: bool = False


def dxy_orbital(coords: jax.Array) -> jax.Array:
    """(Arbitrary) function that looks like a 3dxy orbital."""
    # Scaling coordinates to get shape mostly contained in frame.
    coords *= 8.0
    x_s = coords[..., 0]
    y_s = coords[..., 1]
    r_s = jnp.linalg.vector_norm(coords, axis=-1)
    return jnp.exp(-r_s / 2) * x_s * y_s


def _generate_samples(func: ScalarFunction, thetas: jax.Array, phis: jax.Array) -> jax.Array:
    """Generate a signal using the given sampling angles.

    Axes will correspond to `(theta, phi)`.
    """
    x_s = jnp.sin(thetas)[:, None] * jnp.cos(phis)[None, :]
    y_s = jnp.sin(thetas)[:, None] * jnp.sin(phis)[None, :]
    z_s = jnp.cos(thetas)[:, None] + 0 * x_s
    coords = jnp.stack([x_s, y_s, z_s], axis=-1)
[...]

PR #230, opened 2024-09-30
Add stable high-spin transforms (precompute, standard)

CosmoMatt:
This PR wraps the existing Risbo Wigner d-function recursions to generate the necessary precompute matrices for the forward/inverse spherical harmonic/Wigner transforms. 

### Method:
To do this for each $\ell \in [0, L)$ the Wigner d plane for $m,n$ is computed for $\beta = \frac{\pi}{2}$. Then the coefficients $d^{\ell}_{mn}(\beta)$ for all $\lbrace m,n,\beta \rbrace$ are efficiently computed through an IRFFT (see equation 8 of [this](https://arxiv.org/pdf/1110.6298) paper, or discussion 3.1 of this [paper](https://onlinelibrary.wiley.com/doi/epdf/10.1107/S0108767306017478?saml_referrer)).

A single IRFFT is necessary for each $\lbrace \ell, m, n \rbrace$ each of which is of length $\mathcal{O}(L)$ with complexity $\mathcal{O}(L \log L)$ and therefore the complexity here is $\mathcal{O}(NL^3 \log L)$.  Alternatively we can manually perform the recursion for each $\beta$ with complexity $\mathcal{O}(L^4)$. Note that even though N may be $\leq L$ for the Risbo recursion we still need all $N$ unfortunately. Therefore when $N <= L / \log L$ the FFT approach is better, otherwise it's better to compute the recursion manually.

I've added a switch to automatically switch between these modes depending on the parameter configuration. In any case, the same $\mathcal{O}(NL^3 \sim L^4)$ memory overhead still applies.

Further, more efficient, alternatives will be added in a subsequent PR.


CosmoMatt:
@jasonmcewen I've now reworked the existing functions which precompute the necessary Wigner d-function elements to switch between recursion depending on the users configuration.

Ultimately, a user should now be able to run the precompute harmonic and Wigner transforms at arbitrary spins and recover machine precision (except for HEALPix sampling of course, but that is another issue entirely...)

@ElisR hopefully this should solve your immediate problems, at least up to L = N = 128. I'll pick up the more memory efficient routines in a separate PR.

s2fft/precompute_transforms/construct.py
```diff
@@ -53,27 +60,102 @@ def spin_spherical_kernel(
             "Reality acceleration only supports spin 0 fields. "
             + "Defering to complex transform."
```
jasonmcewen, line 61: Typo "Defering: -> "Deferring".  Check here and throughout.

s2fft/precompute_transforms/construct.py
```diff
@@ -53,27 +60,102 @@ def spin_spherical_kernel(
...
+        mode = "risbo"
+        warn(
+            "The Price-McEwen can become unstable for spins >= 5. "
+            + "Defaulting to slower but more stable Risbo recursions."
+        )
+
+    if mode.lower() == "auto":
+        # This mode automatically determines which recursion is best suited for the
+        # current parameter configuration.
+        mode = "risbo" if abs(spin) >= 5 else "price_mcewen"
```
jasonmcewen, line 73: Can we make the 5 a global variable that we pick up whenever we do this check.  Then if we want to change this we just need to make an update one place in the code.

s2fft/precompute_transforms/construct.py
```diff
@@ -53,27 +60,102 @@ def spin_spherical_kernel(
             "Reality acceleration only supports spin 0 fields. "
             + "Defering to complex transform."
         )
+    if mode.lower() == "price_mcewen" and abs(spin) >= 5:
```
jasonmcewen, line 63: I'm not sure we should override the users selection.  If they select auto we're free to choose the recursion but otherwise we should follow what they set.  Users never read warnings, so I don't think they would realise the recusion has been changed.  

If we are in the setting where we know price-mcewen recursion is going to be unstable, then we should throw a wanring.  Or perhaps we should throw and error and simply stop execution?  What do you think we should generally do?

s2fft/precompute_transforms/construct.py
```diff
@@ -53,27 +60,102 @@ def spin_spherical_kernel(
...
+        dl = np.zeros((len(thetas), L, m_dim), dtype=np.float64)
+        for t, theta in enumerate(thetas):
+            for el in range(abs(spin), L):
+                dl[t, el] = recursions.turok.compute_slice(
+                    theta, el, L, -spin, reality
+                )[m_start_ind:]
+                dl[t, el] *= np.sqrt((2 * el + 1) / (4 * np.pi))
+
+    # Calculate Wigner d-function elements through the Risbo recursion.
+    elif mode.lower() == "risbo":
```
jasonmcewen, line 99: After this `elif` we should have a final `else` that catches if an unsupported recursions is provided.

s2fft/precompute_transforms/construct.py
```diff
@@ -53,27 +60,102 @@ def spin_spherical_kernel(
...
+            )
+            for el in range(L):
+                delta = recursions.risbo.compute_full_vect(delta, thetas, L, el)
+                dl[:, el] = delta[:, m_start_ind:, L - 1 - spin]
+
+        # MW, MWSS, and DH sampling ARE uniform in theta therefore CAN be calculated
+        # using the Fourier decomposition of Wigner d-functions.
+        # - The complexity of this approach is O(L^3LogL).
+        # - This approach is stable for arbitrary abs(spins) <= L.
+        if sampling.lower() in ["mw", "mwss", "dh"]:
```
jasonmcewen, line 119: This should be `elif`.  Also we should have a final `else` to catch any unsupported sampling schemes.

s2fft/precompute_transforms/construct.py
```diff
@@ -44,6 +47,10 @@ def spin_spherical_kernel(
 
         using_torch (bool, optional): Desired frontend functionality. Defaults to False.
 
+        mode (str, optional): Recursion to adopt. Supported recursion schemes include
+            {"auto", "price_mcewen", "risbo"}. Defaults to "auto" which will detect the
```
jasonmcewen, line 51: I would perhaps made the settings `price-mcewen`, i.e. hypenate.  That more closely follows how this would be referenced in text.

s2fft/precompute_transforms/construct.py
```diff
@@ -193,39 +369,103 @@ def wigner_kernel(
         forward (bool, optional): Whether to provide forward or inverse shift.
             Defaults to False.
 
+        mode (str, optional): Whether to use FFT approach or manually compute each element.
```
jasonmcewen, line 372: Mode is good here since we're not selecting recursion but the underlying compute mode.  However, let's switch `manual` to `direct` (`manual` suggests some kind of manual computation to me rather than just computing this directly, which I think is also the terminology sometimes used elsewhere).

s2fft/precompute_transforms/construct.py
```diff
@@ -193,39 +369,103 @@ def wigner_kernel(
...
+            {"auto", "manual", "irfft"}. Defaults to "auto" which will detect the
```
jasonmcewen, line 373: The docstring says "irfft" but it seems "fft" is used in the code?  "fft" as used in the code is probably better since it less technical.  

I'm surpised tests haven't caught this.  Can you also add a test that would catch this?

s2fft/precompute_transforms/construct.py
```diff
@@ -193,39 +369,103 @@ def wigner_kernel(
...
+            most appropriate recursion given the parameter configuration.
+
         using_torch (bool, optional): Desired frontend functionality. Defaults to False.
 
     Returns:
         np.ndarray: Transform kernel for Wigner transform.
     """
+    if mode.lower() == "fft" and sampling.lower() not in ["mw", "mwss", "dh"]:
```
jasonmcewen, line 381: Again, I don't think we should switch the mode if auto is not selected.  I would just fail with an error here.  Then the user can correct their settings and rerun.

s2fft/precompute_transforms/construct.py
```diff
@@ -193,39 +369,103 @@ def wigner_kernel(
...
     dl = np.zeros((n_dim, len(thetas), L, 2 * L - 1), dtype=np.float64)
-    for n in range(n_start_ind - N + 1, N):
-        for t, theta in enumerate(thetas):
-            for el in range(abs(n), L):
-                ind = n if reality else N - 1 + n
-                dl[ind, t, el] = recursions.turok.compute_slice(theta, el, L, n, False)
 
+    # GL and HP sampling ARE NOT uniform in theta therefore CANNOT be calculated
+    # using the Fourier decomposition of Wigner d-functions. Instead they must be
+    # manually bootstrapped from the recursion.
```
jasonmcewen, line 413: Let's not say bootsrapped.  That has a precise statistical meaning.  I would just say "directly computed"...

s2fft/precompute_transforms/construct.py
```diff
@@ -266,57 +509,110 @@ def wigner_kernel_jax(
...
 
+        mode (str, optional): Whether to use FFT approach or manually compute each element.
+            {"auto", "manual", "irfft"}. Defaults to "auto" which will detect the
+            most appropriate recursion given the parameter configuration.
+
     Returns:
         jnp.ndarray: Transform kernel for Wigner transform.
     """
+    if mode.lower() == "fft" and sampling.lower() not in ["mw", "mwss", "dh"]:
+        mode = "manual"
```
jasonmcewen, line 520: Again, don't override mode.

jasonmcewen (changes requested):
Is all looking great @CosmoMatt!  

A few comments but they mostly relate to very minor clarity related things.  Note that for some of these I've only highlighted the first instance, so please be sure to address all related instances.  Thanks! 

One question.  Do we only support Risbo for the Wigner case then?  I don't recall all the details of what we discussed the other day.  Was there no setting where McEwen-Price was better?  Perhaps not actually.

And test coverage seems to be stopping the merge.  Is looks like the new code has good coverage so not exactly sure what's going on there?

ElisR:
Hi @CosmoMatt, this is looking very promising, thanks!

I ran my minimal example on this branch (after changing which API function I called) and the results look excellent, at least on CPU.

Old figure from #209:

broken_s2fft_inverse_wigner

New figure generated with `dec8cb1e`:

fixed_s2fft_inverse_wigner

CosmoMatt:
@ElisR Ok great that it seems to have solved your immediate problem! If there's any further functionality that you'd like to see (or think would be useful) in the package, let us know and we can add it to our board.

@jasonmcewen I'll pick up the suggested changes when I next get a chance, then we can merge before implementing the various memory efficient versions of the Risbo transforms.

s2fft/recursions/risbo.py
```diff
@@ -116,3 +118,105 @@ def _arg_checks(dl: np.ndarray, beta: float, L: int, el: int):
     assert 0 <= el < L  # Should be < not <= once have init routine.
     assert dl.shape[0] == dl.shape[1] == 2 * L - 1
     assert 0 <= beta <= np.pi
+
+
+def compute_full_vect(
+    dl: np.ndarray, beta: np.ndarray, L: int, el: int
+) -> np.ndarray:
+    r"""Compute Wigner-d at argument :math:`\beta` for full plane using
```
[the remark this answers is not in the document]
CosmoMatt: Updated docstrings and switched to _vectorised

s2fft/precompute_transforms/construct.py
CosmoMatt: Introduced a global variable PM_MAX_STABLE_SPIN which is called throughout.

s2fft/precompute_transforms/construct.py
CosmoMatt: @jasonmcewen this is the same behaviour we adopt for if the user provides reality=True for spin != 0. We warn them that this will cause errors, and then default to reality=False.

Ultimately we need to decide whether we should let the user call the functions in a way that we know for sure a priori will cause errors.

If we just warn the user and don't switch, they will get bad results and then spend time trying to figure it out and eventually switch to Risbo.

If we throw a ValueError pointing to Risbo they will switch to Risbo more quickly.

In either case, the user will have to switch to Risbo, so I don't see the harm in helping them out and switching automatically with a warning?

s2fft/precompute_transforms/construct.py
CosmoMatt: This error is caught earlier in lower level functions to compute theta values. No need to duplicate higher up in the code I think?

s2fft/precompute_transforms/construct.py
CosmoMatt: This is just a docstring difference so there isn't a way to test this. The code is otherwise consistent, and I've now updated the docstring to "fft" to keep things simple.

s2fft/precompute_transforms/construct.py
CosmoMatt: I still think that we shouldn't be intentionally allowing bad configuration, but I've changed this to raise a ValueError instead.

s2fft/precompute_transforms/construct.py
CosmoMatt: changed to calculated.

s2fft/precompute_transforms/construct.py
CosmoMatt: changed to raise ValueError instead.

s2fft/precompute_transforms/construct.py
CosmoMatt: Changed to direct.

s2fft/precompute_transforms/construct.py
CosmoMatt: I've changed the behaviour to just raise a ValueError and updated tests accordingly.

s2fft/precompute_transforms/construct.py
```diff
@@ -58,19 +62,19 @@ def spin_spherical_kernel(
...
+            + "Defaulting to complex transform."
         )
-    if mode.lower() == "price_mcewen" and abs(spin) >= 5:
-        mode = "risbo"
-        warn(
-            "The Price-McEwen can become unstable for spins >= 5. "
-            + "Defaulting to slower but more stable Risbo recursions."
+    if recursion.lower() == "price-mcewen" and abs(spin) >= PM_MAX_STABLE_SPIN:
+        raise ValueError(
+            f"The Price-McEwen can become unstable for spins >= {PM_MAX_STABLE_SPIN}."
```
jasonmcewen, line 71: Typo: "Price-McEwen recursion", i.e. recursion missing.

tests/test_spherical_precompute.py
```diff
@@ -14,7 +14,9 @@
 sampling_to_test = ["mw", "mwss", "dh", "gl"]
 reality_to_test = [True, False]
 methods_to_test = ["numpy", "jax", "torch"]
-recursions_to_test = ["price_mcewen", "risbo", "auto"]
+recursions_to_test = ["price-mcewen", "risbo", "auto"]
+
+PM_MAX_STABLE_SPIN = 5
```
jasonmcewen, line 20: This should be picked up from the main library not redefined in the tests.

s2fft/precompute_transforms/construct.py
```diff
@@ -10,6 +10,10 @@
 from s2fft import recursions
 from warnings import warn
 
+# Maximum spin number at which Price-McEwen recursion is sufficiently accurate.
+# For spins > PM_MAX_STABLE_SPIN one should default to the Risbo recursion.
+PM_MAX_STABLE_SPIN = 5
```
jasonmcewen, line 14: We call this PM_MAX_STABLE_SPIN but test are >=, so we're throwing errors for PM_MAX_STABLE_SPIN.  Tests should probably be >, not >=?

jasonmcewen (approved):
@CosmoMatt I caught a couple of very minor typos.  Can you correct these and then merge?  I've approved the PR already so no need for me to take another look.  Please just merge when you've addressed these very minor outstanding points.

s2fft/precompute_transforms/construct.py
CosmoMatt: I agree that this isn't the most clear, we are actually already being quite conservative here (we have stability up to closer to abs(spin) ~ 8,10 or so).

In any case I've updated the checks to be greater than and incremented the global variable to 6.

PR #233, opened 2024-10-07
Set up Ruff for linting and formatting

### LINKED ISSUE #205: Decide on a code style convention and set-up automatic formatting
We don't currently specify the code style convention we want the package code to adhere to. Ideally we should settle on something and document this to make it easy for contributors to know what to target when proposing changes and to ensure we are consistent across the repository to improve readability. This has come up as part of @ASKabalan's contributions in #204.

Currently I think we are mainly following [PEP8 guidelines](https://peps.python.org/pep-0008/) and this is probably the most commonly used convention in open-source Python projects so I would personally suggest adopting this. Alternatives include the [Google Python style guide](https://google.github.io/styleguide/pyguide.html).

Related to this, to simplify keeping to a standard style, it would be good to adopt a specific formatter which enforces our chosen style conventions - as an additional check as part of the CI workflows and/or by setting up [`pre-commit` hooks](https://pre-commit.com/) to automatically run the formatter on any commits. My suggestion for a formatter would be [the Ruff formatter](https://docs.astral.sh/ruff/formatter/) as its very performant so minimizes overhead when formatting and uses the same formatting rules as [Black ](https://github.com/psf/black) which is widely used. 

Tagging @jasonmcewen and @CosmoMatt for their views.

### LINKED ISSUE #216: Set up linting checks
Alongside setting up automated formatting (#205) we should also decide on a [linter](https://en.wikipedia.org/wiki/Lint_(software)) and set up a workflow to automate checks on pull-requests and pushes to main. I would recommend [ruff](https://github.com/astral-sh/ruff) as the linter given its performance. We will need to decide on which [linting rules to select](https://docs.astral.sh/ruff/linter/#rule-selection) - ruff's [default configuration](https://docs.astral.sh/ruff/configuration/) is probably a reasonable starting point though we may want to enable some additional rules and/or disable any which we decide we are going to take a different convention. Depending on how many changes are needed to the codebase to get the checks passing initially, we may also want to take a staged approach of initially enabling a smaller subset of rules, getting the code to pass these and adding in additional rules later.

As with autoformatting, I would say having the linting set up as [a `pre-commit` hook ](https://docs.astral.sh/ruff/integrations/#pre-commit) would have the advantage of catching any changes which fail the checks early (when committing) and we can also re-use the pre-commit checks within a GitHub Actions workflow to ensure consistency / avoid redundancy in set up.

matt-graham:
Resolves #216 
Resolves #205 

Adds configuration for Ruff to `pyproject.toml` file, enabling a relatively small set of rules for now (based on default configuration plus some trial and error of enabling additional rule sets for which there was either automatic fixes available or only a few violations requiring manual fixes).

All current violations of Ruff linting rules in current code are fixed with a mix of automatic fixes and some manual fixes. Code has also been formatted automatically with Ruff formatter.

This pull-request also sets up a [`pre-commit`](https://pre-commit.com/) configuration for installing pre-commit hooks for automatically running the Ruff linter and formatter on each commit and also adds a GitHub Actions workflow for checking these pre-commit linting checks pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
-[![image](https://github.com/astro-informatics/s2fft/actions/workflows/tests.yml/badge.svg?branch=main)](https://github.com/astro-informatics/s2fft/actions/workflows/tests.yml)
-[![image](https://codecov.io/gh/astro-informatics/s2fft/branch/main/graph/badge.svg?token=7QYAFAAWLE)](https://codecov.io/gh/astro-informatics/s2fft)
-[![image](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
-[![image](https://badge.fury.io/py/s2fft.svg)](https://badge.fury.io/py/s2fft)
-[![image](http://img.shields.io/badge/arXiv-2311.14670-orange.svg?style=flat)](https://arxiv.org/abs/2311.14670)<!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
+[![Tests status](https://github.com/astro-informatics/s2fft/actions/workflows/tests.yml/badge.svg?branch=main)](https://github.com/astro-informatics/s2fft/actions/workflows/tests.yml)
+[![Linting status](https://github.com/astro-informatics/s2fft/actions/workflows/linting.yml/badge.svg?branch=main)](https://github.com/astro-informatics/s2fft/actions/workflows/linting.yml)
+[![Documentation status](https://github.com/astro-informatics/s2fft/actions/workflows/docs.yml/badge.svg?branch=main)](https://github.com/astro-informatics/s2fft/actions/workflows/docs.yml)
+[![Codecov](https://codecov.io/gh/astro-informatics/s2fft/branch/main/graph/badge.svg?token=7QYAFAAWLE)](https://codecov.io/gh/astro-informatics/s2fft)
+[![MIT License](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
+[![PyPI package](https://badge.fury.io/py/s2fft.svg)](https://badge.fury.io/py/s2fft)
+[![arXiv](http://img.shields.io/badge/arXiv-2311.14670-orange.svg?style=flat)](https://arxiv.org/abs/2311.14670)<!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
 [![All Contributors](https://img.shields.io/badge/all_contributors-11-orange.svg?style=flat-square)](#contributors-)
 <!-- ALL-CONTRIBUTORS-BADGE:END --> 
 [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/astro-informatics/s2fft/blob/main/notebooks/spherical_harmonic_transform.ipynb)
--- a/s2fft/logs.py
+++ b/s2fft/logs.py
@@ -1,13 +1,16 @@
-import os
-import logging.config
 import logging
+import logging.config
+import os
 from pathlib import Path
+
 import yaml
+
 import s2fft
 
 
 def setup_logging(custom_yaml_path: str = None):
-    """Initialise and configure logging.
+    """
+    Initialise and configure logging.
 
     Should be called at the beginning of code to initialise and configure the
     desired logging level. Logging levels can be ints in [0,50] where 10 is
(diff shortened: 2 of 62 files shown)


tests/test_wigner_samples.py
```diff
@@ -59,7 +60,7 @@ def test_elmn2ind(s2fft_to_so3_sampling, L: int, N: int, sampling: str):
 
 @pytest.mark.parametrize("L", L_to_test)
 @pytest.mark.parametrize("N", N_to_test)
-def test_so3_samples(flmn_generator, L: int, N: int):
+def test_so3_samples_2(flmn_generator, L: int, N: int):
```
matt-graham, line 63: @CosmoMatt This test function has the same name as another function defined earlier in file and so was shadowing this - I have just added a `_2` suffix for now but if only one of these two functions is required we should remove one, or give more useful separate names.

s2fft/recursions/turok.py
```diff
@@ -227,41 +235,43 @@ def compute_quarter_slice(
...
+    Compute the left quarter triangle of the Wigner-d matrix via Turok & Bucher
     recursion.
 
     Args:
         dl (np.ndarray): Wigner-d matrix slice to populate (shape: 2L-1, 2L-1).
 
         beta (float): Polar angle in radians.
 
-        l (int): Harmonic degree of Wigner-d matrix.
+        ell (int): Harmonic degree of Wigner-d matrix.
```
matt-graham, line 248: Ruff flagged this variable name as ambiguous (https://docs.astral.sh/ruff/rules/ambiguous-variable-name/) so I manually changed from `l` to `ell` here and in another function below.

s2fft/precompute_transforms/construct.py
```diff
@@ -342,10 +350,11 @@ def healpix_phase_shifts(
 
     Returns:
         np.ndarray: Vector of phase shifts with shape :math:`[thetas, 2L-1]`.
+
     """
     thetas = samples.thetas(L, "healpix", nside)
     phase_array = np.zeros((len(thetas), 2 * L - 1), dtype=np.complex128)
-    for t, theta in enumerate(thetas):
+    for t in range(len(thetas)):
```
matt-graham, line 357: `theta` unused here so I just changed to a loop over `range(len(thetas))`

s2fft/logs.py
```diff
@@ -29,7 +32,7 @@ def setup_logging(custom_yaml_path: str = None):
         path = Path(custom_yaml_path)
     if not path.exists():
         raise ValueError(f"Logging config path {path} does not exist.")
-    with open(path, "rt") as f:
+    with open(path) as f:
```
matt-graham, line 35: `"rt"` corresponds to the default `mode` for `open` (https://docs.astral.sh/ruff/rules/redundant-open-modes/#pyupgrade-up)

README.md
```diff
@@ -1,8 +1,10 @@
...
+[![Tests status](https://github.com/astro-informatics/s2fft/actions/workflows/tests.yml/badge.svg?branch=main)](https://github.com/astro-informatics/s2fft/actions/workflows/tests.yml)
+[![Linting status](https://github.com/astro-informatics/s2fft/actions/workflows/linting.yml/badge.svg?branch=main)](https://github.com/astro-informatics/s2fft/actions/workflows/linting.yml)
+[![Documentation status](https://github.com/astro-informatics/s2fft/actions/workflows/docs.yml/badge.svg?branch=main)](https://github.com/astro-informatics/s2fft/actions/workflows/docs.yml)
+[![Codecov](https://codecov.io/gh/astro-informatics/s2fft/branch/main/graph/badge.svg?token=7QYAFAAWLE)](https://codecov.io/gh/astro-informatics/s2fft)
+[![MIT License](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
+[![PyPI package](https://badge.fury.io/py/s2fft.svg)](https://badge.fury.io/py/s2fft)
+[![arXiv](http://img.shields.io/badge/arXiv-2311.14670-orange.svg?style=flat)](https://arxiv.org/abs/2311.14670)<!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
```
matt-graham, line 7: I changed the alt-text fields for the badges here to be more descriptive rather than the previous `image` that was used for a lot of them

tests/test_wigner_samples.py
CosmoMatt: good catch!

s2fft/precompute_transforms/construct.py
CosmoMatt: Nice, originally I suspect we were using theta here but then pulled this outside the loop.

README.md
CosmoMatt: Nice update!

matt-graham:
> Addressed each conversation individually. All looks good to me, so long as CI passes fine to merge.

Thanks @CosmoMatt for the quick review! Linting and test workflows both passing here so will merge this now.

PR #238, opened 2024-10-18
add stable forward/inverse memory efficient Wigner transforms

CosmoMatt:
This PR extends the Risbo precompute Wigner transform to the setting where we instead **ONLY** compute the Wigner d coefficients for $\theta = \beta = \pi/2$ which we denote $\Delta^l_{mn} = d^l_{mn}(\pi/2)$. The coefficients for colatitudes other than $\pi/2$ are accessed implicitly at run-time through the Fourier relation outlined by this [McEwen et al 2016](https://arxiv.org/pdf/1508.03101) with further details in this [McEwen et al 2011](http://www.jasonmcewen.org/papers/fssht.pdf).

Ultimately this algorithm costs an extra FFT at run-time but reduces the memory overhead from $\mathcal{O}(NL^3)\sim\mathcal{O}(L^4)$ to $\mathcal{O}(L^3)$. A further reduction in memory by a factor of 8 can be achieved by exploiting the various symmetries of $\Delta^l_{mn}$, however this has yet to be implemented.

As the Risbo recursion is extremely stable we can also consider reduced precision computation (f32, f16, f8) which should recover order ($10^{-8}, 10^{-4}, 10^{-2}$) respectively with corresponding memory reduction and acceleration by a factor of ($2, 4, 8$) respectively.

> [!TIP]
> When building flax layers for $SO(3)$ convolutions using this implementation the dominant memory overhead is entirely independent from the number of channels, so the memory requirement for the precompute transform is fixed.

> [!NOTE] 
> With both the memory reductions from aforementioned symmetry relations and from the switched algorithm, the > precompute algorithm should be runnable at $L \sim 2048$ and potentially slightly beyond. However, this has yet to be tested explicitly.

### DIFF
--- a/s2fft/precompute_transforms/fourier_wigner.py
+++ b/s2fft/precompute_transforms/fourier_wigner.py
@@ -7,7 +7,7 @@ from jax import jit
 
 def inverse_transform(
     flmn: np.ndarray,
-    delta: np.ndarray,
+    DW: np.ndarray,
     L: int,
     N: int,
     reality: bool = False,
@@ -18,7 +18,7 @@ def inverse_transform(
 
     Args:
         flmn (np.ndarray): Wigner coefficients.
-        delta (Tuple[np.ndarray, np.ndarray]): Fourier coefficients of the reduced
+        DW (Tuple[np.ndarray, np.ndarray], optional): Fourier coefficients of the reduced
             Wigner d-functions and the corresponding upsampled quadrature weights.
         L (int): Harmonic band-limit.
         N (int): Azimuthal band-limit.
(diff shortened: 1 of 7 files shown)


s2fft/precompute_transforms/fourier_wigner.py
```diff
@@ -193,14 +220,17 @@ def forward_transform(
     x = np.fft.ifft(x, axis=1, norm="forward")
 
     # PERFORM QUADRATURE CONVOLUTION AS FFT REWEIGHTING IN REAL SPACE
-    x = np.einsum("nbm,b->nbm", x, delta[1])
+    # NB: Our convention here is conjugate to that of SSHT, in which
+    # the weights are conjugate but applied flipped and therefore are
+    # equivalent. To avoid flipping here he simply conjugate the weights.
```
jasonmcewen, line 225: Typo: he -> we

jasonmcewen (approved):

PR #241, opened 2024-11-13
Iterative refinement support for JAX and NumPy forward (spherical) transform implementations

### LINKED ISSUE #139: Support HEALPix iterations to improve accuacy of transforms

matt-graham:
Should resolve #139 

Adds support for using `iter` keyword argument to `s2fft.forward` when using `method = "numpy"` and `method = "jax"`, using an [iterative refinement](https://en.wikipedia.org/wiki/Iterative_refinement) approach equivalent to that implemented in `healpy` to improve accuracy of forward transform (in terms of acting as the inverse of the inverse transform). 

Currently this is only added for spherical transform, and is available for all sampling types, even though in practice probably only useful for HEALPix (though `iter=1` does seem to give a small gain in accuracy for other sampling types).

### DIFF
--- a/s2fft/base_transforms/spherical.py
+++ b/s2fft/base_transforms/spherical.py
@@ -1,3 +1,4 @@
+from functools import partial
 from warnings import warn
 
 import numpy as np
@@ -5,7 +6,7 @@ import numpy as np
 from s2fft import recursions
 from s2fft.sampling import s2_samples as samples
 from s2fft.utils import healpix_ffts as hp
-from s2fft.utils import quadrature, resampling
+from s2fft.utils import iterative_refinement, quadrature, resampling
 
 
 def inverse(
@@ -138,6 +139,7 @@ def forward(
     nside: int = None,
     reality: bool = False,
     L_lower: int = 0,
+    iter: int = 0,
 ) -> np.ndarray:
     r"""
     Compute forward spherical harmonic transform.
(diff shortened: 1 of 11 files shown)


CosmoMatt:
@matt-graham looks great, one thing that would also be good to add is the same iterations but for the precompute versions of the transforms.

jasonmcewen:
Good point @CosmoMatt (re precompute).  We could either do that in the current PR if quick or else do it in a next one.

matt-graham:
@CosmoMatt - just to check by precompute versions, do you mean the `forward` and `inverse` functions in `s2fft/precompute_transforms/spherical.py`?

Also, with regards to the `precomps` argument to the `forward` and `inverse` functions in `s2fft/transforms/spherical.py` - at the moment the current implementation passes whatever value for `precomps` is passed to the top level `forward` call to both the underlying forward and inverse function calls. With `precomps = None` this works fine, but thinking about it I think this will give incorrect results if `precomps != None` as presumable the precomputed values for forward and inverse transforms differ? Also as we are making multiple calls to the forward and inverse transforms when `iter > 0` presumably it is always worth pre-computing in this situation?

matt-graham:
- [ ] Add tests checking base transform with `iter > 0` against `healpy` + check reconstruction for large `n_iter` close to zero

PR #244, opened 2024-11-18
Correct `healpix_forward` derivatives and add support for forward and higher order autodiff 

### LINKED ISSUE #243: Derivatives for `healpix_forward` are incorrrect
I believe the derivatives defined for the `s2fft.transforms.c_backend_spherical.healpy_forward` function with `custom_vjp` are incorrect. As a minimal reproducing example:

```Python
import s2fft
from s2fft.transforms import c_backend_spherical as c_sph
import numpy as np
import jax
from jax.test_util import check_grads
jax.config.update("jax_enable_x64", True)

L = 32
nside = L // 2
method = "jax"
sampling = "healpix"
reality = True
rng = np.random.default_rng(23457801234570)
flm = s2fft.utils.signal_generator.generate_flm(rng, L, reality=reality)
f = s2fft.inverse(flm, L, nside=nside, sampling=sampling, method=method, reality=reality)
```

Running `check_grads` on `c_sph.healpy_inverse` to check gradients against numerical finite differencing completes without any error

```Python
check_grads(lambda x: c_sph.healpy_inverse(x, L, nside), (flm,), modes=("rev",), order=1)
```

Running the same on `c_sph.healpy_forward` however gives an `AssertionError`

```Python
check_grads(lambda x: c_sph.healpy_forward(x, L, nside, iter=0), (f,), modes=("rev",), order=1)
```

outputting

```Python
AssertionError: 
Not equal to tolerance rtol=1e-05, atol=1e-05
VJP cotangent projection
Mismatched elements: 1 / 1 (100%)
Max absolute difference: 1.84452973
Max relative difference: 1.69768919
 x: array(-2.931024)
 y: array(-1.086494)
```

Somewhat confusingly, running `check_grads` instead on specific scalar functions constructed using `c_sph.healpy_forward` does pass without error:

```Python
check_grads(lambda x: abs(c_sph.healpy_forward(x, L, nside, iter=0)).sum(), (f,), modes=("rev",), order=1)
```

Notice in all cases `iter` is fixed to zero so this is not due to using iterative refinement steps. While HEALPix sampling does not exhibit a sampling theorem and so round-tripping through the forward and backward ('inverse') transforms will introduce an error (which iterative refinement in the forward transform can reduce), this only affects the accuracy of the linear operator corresponding to the forward transform being the inverse of the linear operator corresponding to the backward transform. 

Distinct from this property, is that the linear operator represented by the forward transform is a (scaled and conjugated) transposition of the linear operator represented by the backward transform (and vice-versa), and this property holds exactly (modulo floating point error) for the HEALPix forward and backward transforms. Importantly it is this transposition property that is required for implementation of the derivative rules and so the incorrectness of the derivatives in the current implementation cannot be explained by the error in the inverse relationship.

Let the matrix represented by HEALPix forward spherical transform (`map2alm`) be $F \in \mathbb{C}^{m \times n}$ and the matrix represented by HEALPix inverse (backward) spherical transform (`alm2map`) be $B \in \mathbb{C}^{n \times m}$, with a harmonic bandlimit $\ell$ in both cases and HEALPix resolution parameter $r$ and $n = 12r^2$, $m = \ell(\ell + 1) / 2$. That is

$$\texttt{map2alm}(x) = F x \quad\text{and}\quad \texttt{alm2map}(y) = B y.$$

We can construct the matrices numerically by mapping the standard basis vectors through the `map2alm` and `alm2map` functions:

```Python

F = np.stack(
    list(
        map(
            lambda e: healpy.map2alm(e, lmax=L - 1),
            np.identity(12 * nside**2, dtype=float),
        )
    ),
    1,
)

# As alm2map argument is complex need to use both real and imaginary basis vectors

B = np.stack(
    list(
        map(
            lambda e: healpy.alm2map(e, nside=nside, lmax=L - 1),
            np.identity(flm_hp.shape[0], dtype=complex),
        )
    ),
    1,
) - 1j * np.stack(
    list(
        map(
            lambda e: healpy.alm2map(e, nside=nside, lmax=L - 1),
            1j * np.identity(L * (L + 1) // 2, dtype=complex),
        )
    ),
    1,
)
```

We then have the relationships

$$ B = F^H D \quad\text{and}\quad F = D^{-1} B^H$$

where $X^H= \textrm{conj}(X)^T$ (that is the conjugate / Hermitian transpose) and $D$ is a $n \times n$ diagonal matrix with the first $\ell$ entries along the diagonal equal to $3r^2 / \pi$ and the remaining $n - \ell = \ell(\ell - 1) / 2$ diagonal entries equal to $6r^2/\pi$.

Numerical verification:

```Python
D = np.diag([1] * L + [2] * (L * (L - 1) // 2)) *  (3 * nside**2) / np.pi
np.allclose(B, F.T.conj() @ D)
D_inv = np.diag(1 / D.diagonal())
np.allclose(D_inv @ B.T.conj(), F)
```
As `map2alm` and `alm2map` are both linear, their Jacobian vector product (JVP) functions are just the original maps:

$$
  \mathsf{jvp}(\texttt{map2alm})(x)(v) = \partial(x \mapsto F x) v = F v = \texttt{map2alm}(v)
$$
$$
  \mathsf{jvp}(\texttt{alm2map})(y)(v) = \partial(y \mapsto B y) v = B v = \texttt{alm2map}(v)
$$

For the vector Jacobian product (VJP) functions we have

$$
\begin{aligned}
\mathsf{vjp}(\texttt{map2alm})(x)(v) 
&= v^T \partial(x \mapsto F x) \\
&= v^TF \\ 
&= (F^T v)^T\\
&= \textrm{conj}((\textrm{conj}(F)^T\, \textrm{conj}(v)))^T\\
&= (F^H D D^{-1} \,\textrm{conj}(v))^H\\
&= (B D^{-1} \textrm{conj}(v))^H\\
& = \textrm{conj}(\texttt{alm2map}(D^{-1} \textrm{conj}(v)))
\end{aligned}
$$

$$
\begin{aligned}
\mathsf{vjp}(\texttt{alm2map})(y)(v) 
&= v^T\partial(y \mapsto B y) \\
&= v^TB \\ 
&= (B^T v)^T\\
&= \textrm{conj}((\textrm{conj}(B)^T\textrm{conj}(v)))^T\\
&= (D D^{-1} B^H\textrm{conj}(v))^H\\
&= (D F \textrm{conj}(v))^H\\
& = D  \textrm{conj}(\texttt{alm2map}(\textrm{conj}(v)))
\end{aligned}
$$

The current (correct) VJP definition for `healpy_inverse`  is 

https://github.com/astro-informatics/s2fft/blob/909e6f11c6b29ffcdaef6c5e055a3bc57addf297/s2fft/transforms/c_backend_spherical.py#L282-L297

It's a little difficult to relate this to the above derivation as this includes both the VJP for `healpy.alm2map` but also the VJP for `s2fft.sampling.reindex.flm_2d_to_hp_fast` with `healpy_inverse` defined as the composition of these:

https://github.com/astro-informatics/s2fft/blob/909e6f11c6b29ffcdaef6c5e055a3bc57addf297/s2fft/transforms/c_backend_spherical.py#L271-L273

Broadly though we can see the VJP rule corresponds to something which performs $\textrm{conj}(\texttt{alm2map}(\textrm{conj}(v)))$ plus an elementwise scaling operation corresponding to the application of $D$

The current (I think incorrect) VJP definition for `healpy_forward` is 

https://github.com/astro-informatics/s2fft/blob/909e6f11c6b29ffcdaef6c5e055a3bc57addf297/s2fft/transforms/c_backend_spherical.py#L341-L348

with `healpy_forward` here defined as a composition of `healpy.map2alm` and `s2fft.sampling.reindex.flm_hp_to_2d_fast`

https://github.com/astro-informatics/s2fft/blob/909e6f11c6b29ffcdaef6c5e055a3bc57addf297/s2fft/transforms/c_backend_spherical.py#L331-L332

Here we see there is no elementwise scaling before the application of `alm2map` corresponding to multiplication by $D^{-1}$ in $\textrm{conj}(\texttt{alm2map}(D^{-1} \textrm{conj}(v)))$

matt-graham:
Resolves #243 

Corrects derivative rule for `healpy_forward` function in `s2fft.transforms.c_backend_spherical` and refactors both `healpy_forward` and `healpy_inverse` in terms of two new custom JAX primitives `healpy_map2alm` and `healpy_alm2map` that wrap the corresponding healpy `map2alm` and `alm2map` functions. The `jax.interpreters.ad.deflinear` function is used to mark the primitives as linear operators with custom transposition rules (defined in terms of the primitives themselves). This gives us support for both forward and reverse mode automatic differentiation and also higher order derivatives. 

The corresponding tests for gradient correctness in `tests/test_spherical_custom_grads.py` are also updated to be stricter, removing the kludge to relax `rtol` tolerance for `test_healpix_c_backend_forward_custom_gradients` to `iter` dependent values, testing both forward and reverse mode derivatives and up to second order derivatives, and checking the derivatives of the forward / inverse transforms themselves rather than a scalar function using them  (with `jax.test_util.check_grads` able to check gradients of vector valued functions).

There are a few possible further changes / thoughts:

- Currently this PR exposes `healpy_map2alm` and `healpy_alm2map` as new public functions in the API. This may be useful in some applications where users want to use harmonic coefficients which are ordered with HEALPix (ring ordered) indexing convention without round tripping back and forth to 2D layout. However it does introduce some duplication in to API and docstrings. We could instead just make these functions private for internal use. Alternatively if we wanted to make them more generally useful we could also make their signatures more closely match the corresponding healpy functions - for example using `lmax` rather than `L` (with `lmax = L - 1`, using `maps` instead of `f` and `alms` intead of `flm` for first argument names, not requiring `nside` argument for `healpy_alm2map`.
- At the moment the implementation of `healpy_map2alm` fixes the `iter` argument to `healpy.map2alm` to `iter=0`, with iterative refinement then implemented in a Python loop in the `healpy_forward` function. This has the advantage of not requiring us to manually implement the derivative (transposition) rule for `healpy_map2alm` with `iter > 0`, however means we get slightly poorer performance in the forward pass due to using Python rather than C loop. 
- Currently as we are still calling out to the healpy HEALPix Python API using `jnp.array` and `np.array` to convert to and from JAX / NumPy arrays, the `healpy_alm2map` and `healpy_map2alm` primitives (and so `healpy_inverse` and `healpy_forward`) are not compatible with the `jax.jit` transform (which is also the case in the current implementation). We could instead define a lowering which calls out to the underlying C++ implementations, which would give us JIT support and possibly a small performance boost by avoiding some Python interpreter overhead. 
- The `jax.interpreters.ad.deflinear` function we use here to define the derivative behaviour / custom transposition rule is not documented. It is used [internally in various places](https://github.com/search?q=repo%3Ajax-ml/jax%20deflinear&type=code) and was suggested as a way defining derivatives for linear operators in jax-ml/jax/issues/1530 but as it's not part of documented API, it could be subject to breaking changes. We already to an extent rely on some non-documented JAX internals through the custom CUDA primitive added in #204, so this probably doesn't add much additional maintenance burden, but if we did want to avoid using this undocumented feature we could just use `jax.custom_vjp` similar to the current implementation but using the corrected implementation for the derivative rule of  `healpy_forward`, at the expense of not getting forward mode / higher order derivative support.

### DIFF
--- a/s2fft/transforms/c_backend_spherical.py
+++ b/s2fft/transforms/c_backend_spherical.py
@@ -360,6 +360,11 @@ def _healpy_alm2map_transpose(df: jnp.ndarray, L: int, nside: int) -> tuple:
         * (3 * nside**2)
         / jnp.pi
     )
+    # Scale factor above includes the inverse quadrature weight given by
+    # (12 * nside**2) / (4 * jnp.pi) = (3 * nside**2) / jnp.pi
+    # and also a factor of 2 for m>0 to account for the negative m.
+    # See explanation in this issue comment:
+    # https://github.com/astro-informatics/s2fft/issues/243#issuecomment-2500951488
     return (scale_factors * jnp.conj(healpy_map2alm(jnp.conj(df), L, nside)),)
 
 


CosmoMatt (approved):
Interesting JAX functionality for linear black functions which is good to know. As you say @matt-graham could be subject breaking changes at a later date. At least for now though this looks good to me!

jasonmcewen:
Thanks @matt-graham !

In reponse to your further thoughts (in corresponding order).

1. Indeed, good to expose `healpy_map2alm` and `healpy_alm2map` to avoid the harmonic coefficient conversion.  I think fine to keep to our conventions, e.g. specifying `L` rather than `lmax`, so we are consistent throughout `s2fft`.
2. This sounds good to get greater accuracy.  The performance in speed that we tradeoff for this should be fairly minimal this we're typcailly just considering a handlful of iterations, i.e. typically `iter = 3` and we're unlikely to ever have `iter > 6` or so.
3. I think this is fine as is.  I would expect these overheads to be minimal compared to the actual spherical harmonic transforms, especially for high `L`.  Do you agree?
4. This functionality sounds great!  If I understand correctly, this gives us access to higher order derivates also.  Hopefully they will document the related functionality at some point.

jasonmcewen (approved):

matt-graham:
> Interesting JAX functionality for linear black functions which is good to know. As you say @matt-graham could be subject breaking changes at a later date. At least for now though this looks good to me!

>     4. This functionality sounds great!  If I understand correctly, this gives us access to higher order derivates also.  Hopefully they will document the related functionality at some point.

If  jax-ml/jax#24726  is merged this would provide a public API to define both custom VJP and JVPs so would be a better alternative to current approach here, though not specific to linear case. 

>     3. I think this is fine as is.  I would expect these overheads to be minimal compared to the actual spherical harmonic transforms, especially for high `L`.  Do you agree?

Yes, agreed, I think the overhead will be minimal here so not worth worrying about.

PR #245, opened 2024-11-19
Avoid loops in `s2fft.sampling.reindex` functions to reduce compile and run times

matt-graham:
The index conversion functions in `s2fft.sampling.reindex` currently use Python loops (over `L` iterations) to perform the reindexing operations slice by slice. As these loops are unrolled when JIT compiling this leads to long compile times as `L` gets bigger.

This PR refactors the reindexing functions to instead construct arrays of indices and use these to gather or scatter the relevant entries in single indexing operations, thus avoiding the loops. For the functions converting between HEALPix and 2D harmonic coefficient indexing, we can use the `numpy.triu_indices` function to construct the relevant index arrays. For the maps between 1D and 2D layouts we can use `numpy.where` on a boolean array constructed so that the relevant array elements are `True`.  In both cases the explicitly constructed index arrays add some memory overhead - as they depend only `L` which is static they will be constants at compile time, and in all cases are I believe $\mathcal{O}(L^2)$ in size - however this is no larger than the memory requirements of the harmonic coefficient arrays themselves, and the computational graphs of the resulting compiled functions are much smaller (no longer scaling with `L`) so its unclear if overall the memory requirements are much larger.

I've done some benchmarking of the new implementations here compared to the current implementations and in all cases tested the compile times are significantly lower and the run times comparable or significantly better:

https://gist.github.com/matt-graham/7a2b5b77f51b5301b8910ced176d8a8a

For example for the `flm_2d_to_hp_fast` function, the following plot shows the compile times of the current implementation ("Current"), the proposed implementation ("triu_indices based") and two other alternatives I tried out ("Slice based" and "Nested fori_loop") for various `L` values

flm_2d_to_hp compile time comparison

and the corresponding plot for run times of the compiled functions

flm_2d_to_hp run time comparison

On both compile and run times the proposed "triu_indices based" implementation is significantly quicker than both the current implementation and the other alternatives considered.

EDIT: There was a slight bug in the `flm_hp_to_2d_fast` reimplementation now hopefully fixed. The run times for the new function are now a little slower but the compile time is still much better than current implementation and importantly not quickly growing with `L`.

### DIFF
--- a/s2fft/sampling/reindex.py
+++ b/s2fft/sampling/reindex.py
@@ -8,7 +8,7 @@ from jax import jit
 @partial(jit, static_argnums=(1))
 def flm_1d_to_2d_fast(flm_1d: jnp.ndarray, L: int) -> jnp.ndarray:
     r"""
-    Convert from 1D indexed harmnonic coefficients to 2D indexed coefficients (JAX).
+    Convert from 1D indexed harmonic coefficients to 2D indexed coefficients (JAX).
     
     Note:
         Storage conventions for harmonic coefficients :math:`flm_{(\ell,m)}`, for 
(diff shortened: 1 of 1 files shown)


jasonmcewen (approved):
LGTM!

CosmoMatt:
@matt-graham nice, this was a temporary loop that I never got around to removing!

CosmoMatt (approved):
Nice!

PR #252, opened 2024-12-04
Vectorize signal generator functions

matt-graham:
When running some benchmarks with large `L` I noticed that the signal generation functions in `s2fft.utils.signal_generator` become very slow for large `L` due to the nested for loops.

This switches to using a vectorized implementation. ~~Currently implemented for `generate_flm` only.~~

EDIT: Now implemented for both `generate_flm` and `generate_flmn`.

### DIFF
--- a/s2fft/transforms/otf_recursions.py
+++ b/s2fft/transforms/otf_recursions.py
@@ -80,7 +80,14 @@ def inverse_latitudinal_step(
     half_slices = [el + mm + 1, el - mm + 1]
 
     if precomps is None:
-        precomps = generate_precomputes(L, -mm, sampling, nside, L_lower)
+        precomps = generate_precomputes(
+            L=L,
+            spin=-mm,
+            sampling=sampling,
+            nside=nside,
+            forward=False,
+            L_lower=L_lower,
+        )
     lrenorm, vsign, cpi, cp2, indices = precomps
 
     for i in range(2):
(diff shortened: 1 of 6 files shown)


s2fft/transforms/otf_recursions.py
```diff
@@ -80,7 +80,14 @@ def inverse_latitudinal_step(
...
     if precomps is None:
-        precomps = generate_precomputes(L, -mm, sampling, nside, L_lower)
+        precomps = generate_precomputes(
+            L=L,
+            spin=-mm,
+            sampling=sampling,
+            nside=nside,
+            forward=False,
+            L_lower=L_lower,
+        )
```
matt-graham, line 90: This is same change as fix in #256 to resolve #255. Applied here as otherwise test in `tests/test_signal_generator.py::test_generate_flm` was failing with `L_lower != 0`.

s2fft/utils/signal_generator.py
```diff
@@ -87,26 +153,46 @@ def generate_flmn(
...
+            elif n > 0:
+                # Generate independent complex coefficients for positive n
+                flmn[N - 1 + n, min_el:L, L - 1] = complex_normal(
+                    rng, L - min_el, var=2
+                )
+                # For m = 0, n > 0
+                # flmn[N - 1 - n, el, L - 1] = (-1)**n * flmn[N - 1 + n, el, L - 1].conj
+                flmn[N - 1 - n, min_el:L, L - 1] = (-1) ** n * (
+                    flmn[N - 1 + n, min_el:L, L - 1].conj()
+                )
```
matt-graham, line 174: Compared to previous implementation, here for `m = 0` and `abs(n) > 0` we generate complex coefficients rather than real, as the conjugate symmetry condition does not seem to enforce reality in this case - not sure if this is correct though?

PR #256, opened 2024-12-13
Fix pass through of arguments to `generate_precomputes` in NumPy forward spherical transform

### LINKED ISSUE #255: NumPy inverse spherical transform with `precomps=None` and `L_lower > 0` raising error
Calling NumPy spherical transforms without passing `precomps` gives errors in some cases:

```Python
import numpy as np
import s2fft
rng = np.random.default_rng(1234)
L, L_lower, reality = 8, 1, True
flm = s2fft.utils.signal_generator.generate_flm(rng, L=L, L_lower=L_lower, reality=reality)
f = s2fft.inverse(flm, L=L, L_lower=L_lower, reality=reality)
```

gives error traceback

```
File ~/projects/s2fft/s2fft/transforms/otf_recursions.py:102, in inverse_latitudinal_step(flm, beta, L, spin, nside, sampling, reality, precomps, L_lower)
     95 lamb = (
     96     np.einsum("l,t->tl", el + 1, omc)
     97     + np.einsum("l,t->tl", 2 - L + el, c)
     98     - half_slices[i]
     99 )
    100 lamb = np.einsum("tl,t->tl", lamb, 1 / s)
--> 102 dl_iter[1, :, lind:] = np.einsum(
    103     "l,tl->tl",
    104     cpi[0, lind:],
    105     dl_iter[0, :, lind:] * lamb[:, lind:],
    106 )
    108 # Sum into transform vector 0th component
    109 ftm[:, sind + m_offset] = np.nansum(
    110     dl_iter[0, :, lind:]
    111     * vsign[sind, lind:]
   (...)
    114     axis=-1,
    115 )

ValueError: could not broadcast input array from shape (8,2) into shape (8,1)
```

while

```Python
import numpy as np
import s2fft
rng = np.random.default_rng(1234)
L, L_lower, reality = 8, 1, True
flm = s2fft.utils.signal_generator.generate_flm(rng, L=L, L_lower=L_lower, reality=reality)
precomps = s2fft.generate_precomputes(L, L_lower=L_lower, forward=False)
f = s2fft.inverse(flm, L=L, L_lower=L_lower, reality=reality, precomps=precomps)
```

runs without error.

I believe this is due to the lines

https://github.com/astro-informatics/s2fft/blob/8f6e4d5a8e4601ae99df01694d9e08600cef2549/s2fft/transforms/otf_recursions.py#L82-L83

Looking at the signature of `generate_precomputes`

https://github.com/astro-informatics/s2fft/blob/8f6e4d5a8e4601ae99df01694d9e08600cef2549/s2fft/recursions/price_mcewen.py#L15-L22

it looks like we are passing `L_lower` as the value for the `forward` argument. This means for `L_lower != 0` this evaluates as `True` (therefore generating precomputes for forward transform in inverse transform) and also `L_lower` is assumed to be zero when generating precomputes.

We should probably have some tests to cover `precomps = None` case as it looks like we don't currently.

matt-graham:
Resolves #255

### DIFF
--- a/s2fft/transforms/otf_recursions.py
+++ b/s2fft/transforms/otf_recursions.py
@@ -80,7 +80,14 @@ def inverse_latitudinal_step(
     half_slices = [el + mm + 1, el - mm + 1]
 
     if precomps is None:
-        precomps = generate_precomputes(L, -mm, sampling, nside, L_lower)
+        precomps = generate_precomputes(
+            L=L,
+            spin=-mm,
+            sampling=sampling,
+            nside=nside,
+            forward=False,
+            L_lower=L_lower,
+        )
     lrenorm, vsign, cpi, cp2, indices = precomps
 
     for i in range(2):
--- a/tests/test_spherical_transform.py
+++ b/tests/test_spherical_transform.py
@@ -27,6 +27,7 @@ multiple_gpus = [False, True]
 @pytest.mark.parametrize("method", method_to_test)
 @pytest.mark.parametrize("reality", reality_to_test)
 @pytest.mark.parametrize("spmd", multiple_gpus)
+@pytest.mark.parametrize("use_generate_precomputes", [True, False])
 @pytest.mark.filterwarnings("ignore::RuntimeWarning")
 def test_transform_inverse(
     flm_generator,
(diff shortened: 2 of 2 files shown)


CosmoMatt (approved):
Good catch! A lot of the development for the numpy functions was a little more lax as we only really used this functionality for regression testing of the JAX functions.

PR #257, opened 2024-12-13
Indicate cubic memory overhead in `generate_precomputes` docstring

### LINKED ISSUE #246: NumPy `generate_precomputes` implementation has cubic in L memory overhead
The docstring of the the `s2ftt.recursions.price_mcewen.generate_precomputes` function

https://github.com/astro-informatics/s2fft/blob/7cf80bb87201fbdf39a1531a31b25a3112c1f140/s2fft/recursions/price_mcewen.py#L15-L26

suggests the memory overhead is $\mathcal{O}(L^2)$ but the array allocation in

https://github.com/astro-informatics/s2fft/blob/7cf80bb87201fbdf39a1531a31b25a3112c1f140/s2fft/recursions/price_mcewen.py#L75

is I think $\mathcal{O}(L^3)$ as `ntheta` is I think $\mathcal{O}(L)$.

The JAX version in `s2ftt.recursions.price_mcewen.generate_precomputes_jax` I think avoids this allocation and so is $\mathcal{O}(L^2)$. 

While we probably want people to use the JAX version, I was caught out by this as we are currently using the NumPy `generate_precomps` version in the benchmarks 

https://github.com/astro-informatics/s2fft/blob/7cf80bb87201fbdf39a1531a31b25a3112c1f140/benchmarks/spherical.py#L34
https://github.com/astro-informatics/s2fft/blob/7cf80bb87201fbdf39a1531a31b25a3112c1f140/benchmarks/spherical.py#L74

and I got a out-of-memory error when trying to run benchmarks with `L = 4096` as allocating `log_first_row` would require 1TiB of memory!

Possible fixes
- Just document the current implementation is actually $\mathcal{O}(L^3)$ and switch to using JAX version in benchmarks
- Remove the NumPy version altogether and just always use the JAX implementation (we can just wrap and internally deal with converting from JAX arrays to NumPy arrays)
- Update the NumPy implementation to match the optimization in the JAX version

matt-graham:
Resolves #246

### DIFF
--- a/s2fft/recursions/price_mcewen.py
+++ b/s2fft/recursions/price_mcewen.py
@@ -21,9 +21,11 @@ def generate_precomputes(
     L_lower: int = 0,
 ) -> List[np.ndarray]:
     r"""
-    Compute recursion coefficients with :math:`\mathcal{O}(L^2)` memory overhead.
-    In practice one could compute these on-the-fly but the memory overhead is
-    negligible and well worth the acceleration.
+    Compute recursion coefficients with :math:`\mathcal{O}(L^3)` memory overhead.
+
+    The corresponding JAX implementation in :py:func:`generate_precomputes_jax` has a
+    :math:`\mathcal{O}(L^2)` overhead and so should generally be preferred, particularly
+    for large :math:`L`.
 
     Args:
         L (int): Harmonic band-limit.


CosmoMatt (approved):
Seems reasonable (and correct of course). We really were a little too lax on the numpy development, again the reason (_excuse_) being that our focus was on the JAX side of things!

PR #260, opened 2025-01-21
On-the-fly support for Fourier Wigner transforms with Risbo recursions

CosmoMatt:
Updates the Fourier Wigner transforms to add support for the case in which no precomputes are provided. In this case the appropriate values for the Wigner Fourier coefficients (and quadrature weights for the forward transform) are computed on-the-fly.


jasonmcewen:
To clarify, this uses Risbo recursion.

We discussed refactoring this so Risbo is supported for (spin) spherical harmonic transforms as well as Wigner transforms.  However, it is probably better to consider that in a separate PR, so let's go ahead and merge this.

matt-graham:
Tests on Python 3.10+ failing due to #299 - once #300 is merged we should merge in changes from `main` here and rerun.

matt-graham (approved):

PR #261, opened 2025-01-23
Record numerical error in benchmarks and cost + memory estimates for JAX functions

### LINKED ISSUE #254: Record round-trip errors in benchmarks
Ideally benchmarks should record the round-trip errors through forward-inverse / inverse-forward transforms alongside more performance oriented metrics to give idea of numerical accuracy across different configurations / parameters (particularly if testing at single floating point precision).

matt-graham:
Resolves #254

Updates benchmarks to record round-trip error for forward transforms - that is starting from a set of coefficients and computing a signal by performing inverse transform, computing maximum absolute difference of coefficients recovered using forward transform from signal compared to original reference coefficients.

Also updates benchmarking for JAX functions to use the [static cost and memory analysis available for ahead of time compiled functions](https://jax.readthedocs.io/en/latest/aot.html#debug-information-and-analyses-when-available) to get estimates of operation and memory costs, which should give estimates when running both on CPU and GPU, thus giving some way of tracking memory performance on GPU.

The benchmarks have also been updated to use the standard library `tracemalloc` module to try to trace memory usage when running on a CPU rather than the previous usage of [the `memory_profiler` package](https://pypi.org/project/memory-profiler/) which did give very reliable results. This only gives reasonable values for the NumPy methods (presumably as JAX's memory allocations are transparent to the `tracemalloc` module) but as we don't have the cost / memory analysis statistics for NumPy this is still useful.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -12,7 +12,6 @@ If the [`py-cpuinfo` package](https://pypi.org/project/py-cpuinfo/)
 is installed additional information about CPU of system benchmarks are run on will be
 recorded in JSON output.
 
-
 ## Description
 
 The benchmark scripts are as follows:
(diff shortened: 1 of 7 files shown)


jasonmcewen:
This is great @matt-graham! Can we record mean absolute error of harmonic coefficients as well as max absolute error.

matt-graham:
@jasonmcewen thanks - I've now updated to also record the mean absolute error.

Issue #229, opened 2024-09-27
[Suggestion] : Remove strict requirement on backends.


ASKabalan:
Currently, we have to install pytorch to use the JAX backend of s2fft.

I think it would be a nice to be able to conditionally activate backends depending on the availability of the package

For example I user can use s2FFT using only numpy if he does not have JAX or pytorch
Otherwise he gets a runtime error instead of an import time error

Same for pyssht.
Don't know the best practice to do this, but a try catch around the import + global boolean should do the trick

jasonmcewen:
Thanks for the suggestion @ASKabalan!  At present we only have PyTorch support for the precompute approach but we do plan to add PyTorch support for the on-the-fly transforms as well.  So it would be good to have a nice setup like you're suggesting.

@matt-graham do you have any thoughts on this or know of best practices?

matt-graham:
Hi @jasonmcewen. This is related to what I suggested in #224  - we can easily make all of PyTorch, `pyssht` and `healpy` optional dependencies and guard the imports from these packages within `try: ... except ImportError: ...` logic, either at a module level or within the relevant functions themselves, and raise an informative error if a user tries to use a function for which the relevant optional dependencies are not available. 

I would say dropping hard requirements on all external packages which are not 'core' to the package would be good practice for a library like `s2fft` as it avoids forcing users who only want to rely on a subset of the functionality to install unnecessary dependencies, and as I mentioned in #224 extends the range of systems that users can install the package on. The overhead for those users who wish to use the optional features is minimal as we can add relevant optional dependencies groups so that for example they could just do `pip install s2fft[all]` to install all optional dependencies, or `pip install s2fft[torch]` to just install extra dependencies required for PyTorch support and so on.

matt-graham:
Just noticed suggestion to also apply this to JAX - while we could do this for JAX too, as a lot of the modules (perhaps the majority?) make multiple imports from `jax` package namespace, this would potentially get a bit unwieldy to guard all these imports. What might help with reducing our explicit imports from JAX and PyTorch APIs (and duplication of functions) is to use their support for [the array API](https://data-apis.org/array-api/latest/), by doing something like

```Python
import array_api_compat

...

def spectral_periodic_extension_array_api(fm, L: int):
    xp = array_api_compat.array_namespace(fm)
    nphi = fm.shape[0]
    return xnp.concatenate(
        (
            fm[-xnp.arange(L - nphi // 2, 0, -1) % nphi],
            fm,
[...]

PR #267, opened 2025-01-27
Make pyssht and healpy optional dependencies

### LINKED ISSUE #266: Add support for NumPy v2.0
We currently pin our NumPy dependency to `<2`:
 
https://github.com/astro-informatics/s2fft/blob/876e0901382a0a0b1b585c2e897fbcc2612be188/pyproject.toml#L29-L30

This is problematic for users relying on other packages requiring NumPy 2 or above.

matt-graham:
Related to #229 

Removes `pyssht` and `healpy` from required dependencies for `s2fft` as these are currently only used in tests and in JAX wrapper functions in `s2fft.transforms.c_backend_spherical`. Instead they are now added as test specific dependencies and the  `s2fft.transforms.c_backend_spherical` module updated to try to import these modules when calling the wrapper functions and raise an exception with an informative error message instructing the user they need to install the relevant package to use the wrapper if not already installed.

As the dependency on `pyssht` is what was requiring pinning NumPy to `<2` this resolves #266 as `s2fft` can then be installed with NumPy v2 or above if not installing `pyssht`, although here we still pin the NumPy version in the test dependencies so we are not testing against NumPy v2.

### DIFF
--- a/notebooks/JAX_HEALPix_frontend.ipynb
+++ b/notebooks/JAX_HEALPix_frontend.ipynb
@@ -19,9 +19,9 @@
     "import sys\n",
     "IN_COLAB = 'google.colab' in sys.modules\n",
     "\n",
-    "# Install s2fft and data if running on google colab.\n",
+    "# Install s2fft and healpy if running on google colab.\n",
     "if IN_COLAB:\n",
-    "    !pip install s2fft &> /dev/null"
+    "    !pip install s2fft healpy &> /dev/null"
    ]
   },
   {
--- a/notebooks/JAX_SSHT_frontend.ipynb
+++ b/notebooks/JAX_SSHT_frontend.ipynb
@@ -19,9 +19,9 @@
     "import sys\n",
     "IN_COLAB = 'google.colab' in sys.modules\n",
     "\n",
-    "# Install s2fft and data if running on google colab.\n",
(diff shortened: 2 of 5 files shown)


jasonmcewen:
This is very timely @matt-graham !  I was just today discussing this with someone who was very intersted in dropping some dependencies for ease of intergration in [Euclid](https://en.wikipedia.org/wiki/Euclid_(spacecraft)) pipelines. 

While removing ssht as a dependency is good for now, we should actually update ssht and so3 to support numpy>=2.  This is discussed in [this ssht issue](https://github.com/astro-informatics/ssht/issues/77) and [this so3 issue](https://github.com/astro-informatics/s2let/issues/55).  It seems we just need to update PyPi to pick up new Conan recipes.  Do you know how to do that @matt-graham?

matt-graham:
> While removing ssht as a dependency is good for now, we should actually update ssht and so3 to support numpy>=2. This is discussed in [this ssht issue](https://github.com/astro-informatics/ssht/issues/77) and [this so3 issue](https://github.com/astro-informatics/s2let/issues/55). It seems we just need to update PyPi to pick up new Conan recipes. Do you know how to do that @matt-graham?

No sorry, I have not used Conan previously.

matt-graham:
Discussed this in meeting today and conclusion was to merge this now without further review, so merging.

PR #270, opened 2025-02-12
Fix syntax in `CITATION.cff`

matt-graham:
Some extra quotation marks creeped in to `CITATION.cff` file somehow when I first added which is preventing it from being parsed correctly.


matt-graham:
Merging this without review based on discussion in meeting today.

PR #271, opened 2025-02-13
Clean-up unused files

matt-graham:
- Removes `s2ftt.logs` module as this appears not to be used anywhere and associated tests, dependencies and configuration file.
- Removes some configuration files added in #204 which are not in use.
- Updates reference to `README` in manifest file to point to `.md` file rather than now non-existent `.rst` version and removes reference to non-existent `pytest.ini` file.


jasonmcewen (approved):
LGTM!

PR #272, opened 2025-02-13
Allow changing orientation of benchmark plots

matt-graham:
Adds option for plotting benchmark plots with functions along row axis rather than column axis. Needed this for producing plots for Isambard AI benchmarking results report.


matt-graham:
Merging this without review as per discussion in meeting on 2025-03-13 as this only touches benchmarking plots code not library itself.

PR #275, opened 2025-02-24
Use separate jobs and environment for PyPI and Test PyPI uploading

### LINKED ISSUE #259: Upload to PyPI job in build workflow failing
The `upload_pypi` job in the Actions workflow defined in `.github/workflows/build.yml`, which is triggered on new releases, [is failing](https://github.com/astro-informatics/s2fft/actions/runs/12409174093/job/34645884231) with an error

```
Attestation generation failure:

/github/workspace/dist/s2fft-1.2.0.tar.gz already has a publish attestation: /github/workspace/dist/s2fft-1.2.0.tar.gz.publish.attestation

You're seeing this because the action attempted to generated PEP 740
attestations for its inputs, but failed to do so.
```

Specifically this happens in the _Publish package distribution to PyPI_ step after the previous _Publish package distribution to Test PyPI_ step successfully completes.

I think we are hitting against the issue described in https://github.com/pypa/gh-action-pypi-publish/issues/283#issuecomment-2463268124, specifically that the step publishing to Test PyPI creates an attestation file (used to sign / verify the package upload as far as I can tell), which a subsequent `twine upload` command in the step publishing to PyPI finds and raises an error (as it expects there to be no existing attestation file present).

From the discussion in the issue thread it appears there are a couple of possible resolutions:
- [Disable attestations being generated](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#generating-and-uploading-attestations) on one of two steps (probably the test PyPI) by setting `attestations: false` in `with` block for step.
- Refactor the workflow to use separate jobs for publishing to Test PyPI and PyPI. This appears to be the recommended approach as then they each can be operated in a separate environment with its own restrictions.

matt-graham:
Resolves #259 

Separates current combined Actions workflow job for uploading to both Test PyPI and PyPI in to separate jobs each with their own dedicated environment, to avoid the issue with attestation file from Test PyPI upload causing PyPI upload to subsequently fail, with runs in separate jobs instead being in isolated environments. This also seems to be [the recommended practice ](https://github.com/pypa/gh-action-pypi-publish/issues/283#issuecomment-2463268124).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,13 +74,15 @@ jobs:
           name: cibw-sdist
           path: dist/*.tar.gz
 
-  upload_pypi:
+  upload_test_pypi:
     needs: [build_wheels, build_sdist]
     runs-on: ubuntu-latest
-    environment: pypi
+    environment: test-pypi
     permissions:
       id-token: write
-    if: github.event_name == 'release' && github.event.action == 'published'
+    if: |
+      (github.event_name == 'release' && github.event.action == 'published') || 
+      (github.event_name == 'push' && github.ref == 'refs/heads/main')
     steps:
       - uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


matt-graham:
Merging this without review as per internal discussion as this should only affect release workflow.

PR #277, opened 2025-03-11
Provide PyTorch implementations by wrapping JAX functions

matt-graham:
This PR removes the current manual reimplementation of the precompute transform and some of the utility functions provided by `s2fft` to allow use with PyTorch in favour of wrapping the JAX implementations using JAX and PyTorch's mutual support for the DLPack standard [as outlined by Matt Johnson in this Gist](https://gist.github.com/mattjj/e8b51074fed081d765d2f3ff90edf0e9).

Some local benchmarking suggests there is no performance degradation with this wrapping approach compared to the 'native' implementations beyond the very smallest bandlimits `L` and a potential a small constant factor speedup for larger `L` - see benchmarks results in files below

[precompute-spherical-torch-benchmarks.json](https://github.com/user-attachments/files/19195308/precompute-spherical-torch-benchmarks.json)
[precompute-spherical-torch-wrapper-benchmarks.json](https://github.com/user-attachments/files/19195309/precompute-spherical-torch-wrapper-benchmarks.json)

As all imports from `torch` are after changes in this PR confined to the `s2fft.utils.torch_wrapper` module and the import there is guarded in an `try: ... except ImportError` block this PR also removes `torch` from the required dependencies for the project, with an informative error message being raised when the user tries to use the wrapper functionality without `torch` being installed.

## Todo

- [ ] Add tests for functions introduced in `torch_wrapper` module
- [ ] Decide whether to keep wrapped utility modules `s2fft.utils.quadrature_torch` and `s2fft.utils.resampling_torch`
- [ ] Add wrappers for non-precompute transforms
- [ ] Update documentation to reflect new wider support for PyTorch


s2fft/precompute_transforms/spherical.py
```diff
@@ -518,7 +442,9 @@ def forward_transform_jax(
 
     flm = jnp.zeros(samples.flm_shape(L), dtype=jnp.complex128)
     flm = flm.at[:, m_start_ind:].set(
-        jnp.einsum("...tlm, ...tm -> ...lm", kernel, ftm, optimize=True)
+        jnp.einsum(
+            "...tlm, ...tm -> ...lm", kernel.astype(flm.dtype), ftm, optimize=True
```
matt-graham, line 446: Explicit `dtype` cast here was needed to avoid getting warnings when computing gradients through this function which were causing test failures, which indicated an implicit cast from complex to real types was happening which might be loosing information. I think this was due to the `kernel` argument here being of real type but as `ftm` is complex and the `flm` output of the `einsum` is complex, `kernel` will be implicitly cast to complex as part of `einsum` and in the reverse-pass the derivatives with respect to the `kernel` will therefore be complex even though we only should retain the real valued component.

matt-graham:
As discussed in meeting today marking as this as ready for review and requesting review from @kmulderdas. The remaining todos on exposing torch versions of on-the-fly transforms and also at that point documenting wider torch support can be dealt with an a separate PR.

matt-graham:
> As discussed in meeting today marking as this as ready for review and requesting review from @kmulderdas. The remaining todos on exposing torch versions of on-the-fly transforms and also at that point documenting wider torch support can be dealt with an a separate PR.

I've now added PyTorch wrappers for the on-the-fly versions of spherical / Wigner-d transforms and updated docs and tests accordingly.

kmulderdas (approved):
I've gone over the changes and they look great. I don't see any reason for not merging this PR, although it might be good to have a chat at a later time about the breaking JAX changes ([v0.6.0](https://docs.jax.dev/en/latest/ffi.html)).

PR #278, opened 2025-03-17
Add notebook demonstrating equivalence of equiangular samplings

jasonmcewen:


notebooks/equivalence_equiangular_sampling.ipynb
```diff
@@ -0,0 +1,375 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "id": "de96933f-6b89-4796-bdcd-1eb4959a1a55",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "thetas_mw_fejer1 = (np.arange(0, n) + 0.5) * np.pi / n"
```
matt-graham, line 106: It would be useful to have a reference for this formula or at least some explanation of how it arises.

notebooks/equivalence_equiangular_sampling.ipynb
```diff
@@ -0,0 +1,375 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 11,
+   "id": "5334dba6-f247-45ea-8340-9e13d1321ab9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "thetas_mwss_fejer2 = np.arange(1, n) * np.pi / n"
```
matt-graham, line 187: Again without context unclear where this formula comes from / what is being shown here.

matt-graham (changes requested):
I've added some comments. Ideally I would say if we want to include this as a tutorial notebook there should be a bit more explanatory text about what the various checks / asserts in the notebook are showing and where the expressions for computing the samples points come from. For this to appear in the documentation I believe a corresponding link file needs to be added to [`docs/tutorials`](https://github.com/astro-informatics/s2fft/tree/main/docs/tutorials).

jasonmcewen:
Thanks for your comments @matt-graham.  Agreed this needs some further clarification.  I don't think we should have this as a tutorial notebook as this point (i.e. it shouldn't be rendered in the documentation for now).  We can perhaps extend it in future (once we have the additional sampling schemes supported) and then have a more fully featured tutorial notebook.  Nevertheless, I'll make some clarifications shortly...

dipak1munshi:
Thanks for the invitation to review. I have very straightforward comments.
The notebook looks great. I have some simple comments. Probably not too useful.
1. Adding some references to the papers where the samplings are introduced
can be useful to first-time readers.
2, Some introduction to the notations L, n, .. will be helpful in the notebook.
3. I am finishing debugging the CC quadrature. I will try to do a similar notebook.
But, feel free to ignore if the comments are a bit too trivial. Thanks a lot again!

PR #281, opened 2025-03-20
Update cibuildwheel Action version in build workflow and skip MacOS arm64 PyPy builds

matt-graham:
Hopefully will resolve #280

PR #282, opened 2025-03-20
Set up dependabot to automatically update Actions

matt-graham:
Adds a basic [dependabot configuration](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates) to check weekly for available updates to versions of GitHub Actions used in workflows and open PRs to update.

Should hopefully make it less likely we hit against issues like #281


jasonmcewen (approved):
Again, LGTM although I haven't tested anything.

PR #284, opened 2025-03-25
Explicitly checkout tags in build workflow

### LINKED ISSUE #283: Build workflow not inferring version number correctly
Within the build workflow the `cibuildwheel` jobs are not correctly inferring the version number using `setuptools_scm` from the Git tags due to 
https://github.com/actions/checkout/issues/1471 which means that the tags are not being checked out by the `actions/checkout` step. See for example [this run](https://github.com/astro-informatics/s2fft/actions/runs/13996983287/job/39194483107). 

We need to update the checkout Action configuration to explicitly request tags:

```YAML
- uses: actions/checkout@v4
  with:
    fetch-depth: 0
    fetch-tags: true
```

matt-graham:
Resolves #283

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,12 +51,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-
+        with:
+          fetch-depth: 0
+          fetch-tags: true
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.23.1
         env:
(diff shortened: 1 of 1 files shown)


jasonmcewen (approved):
LGTM although I didn't test anything.

matt-graham:
> LGTM although I didn't test anything.

The build jobs run here on the PR suggest this is working as expected as the wheels and source distribution are getting labelled with the correct version.

PR #290, opened 2025-03-26
Updating Healpix CUDA primitive

ASKabalan:
Adding a few updates

- [x]   Updating to the newest custom call API (API 4) using [FFI](https://docs.jax.dev/en/latest/ffi.html)
- [x]   implementing a grad rule  for healpix cuda FFT
- [ ]  Implementing a Batching rule

A batching rule seems to be very important for two things
Being able to jacrev/ jacfwd
and because in most cases .. the size of a healpix map can fit on a single GPU but sometimes we want to batch the spherical transform

I will be doing that next


ASKabalan:
Hello @matt-graham  @jasonmcewen  @CosmoMatt 

Just a quick PR to wrap up a few stuff

1. Updated the binding API to the newest [FFI](https://docs.jax.dev/en/latest/ffi.html)
2. Added a vmap implementation of the cuda primitive
3. Added a transpose rule which allows jacfwd and jacrev (consequently grad aswell)
4. added more tests https://github.com/astro-informatics/s2fft/blob/ASKabalan/tests/test_healpix_ffts.py#L100
5. Removed two files which are now no longer needed with the FFI API ([kernel helpers](https://github.com/astro-informatics/s2fft/blob/main/lib/include/kernel_helpers.h)) (so maybe they should be removed from the license section)
6. Constrained nanobind to be nanobind >=2.0,<2.6" because of a regression https://github.com/wjakob/nanobind/issues/982

And finally I added [cudastreamhandler](https://github.com/astro-informatics/s2fft/blob/ASKabalan/lib/include/cudastreamhandler.hpp) which is used to split the XLA provided stream for the VMAP lowering (this header is my own work)

There is an issue with building [pyssht](https://github.com/astro-informatics/s2fft/actions/runs/14133039257/job/39597964501?pr=290) not sure that this is my fault

I will check the failing worflows when I get the chance, but in the meantime a review is appreciated

tests/test_healpix_ffts.py
```diff
@@ -92,3 +93,94 @@ def test_healpix_ifft_cuda(flm_generator, nside):
...
+        return ftm
+
+    ftm_stacked = jnp.stack([generate_flm() for _ in range(10)], axis=0)
+    ftm = ftm_stacked[0].real
+
+    def healpix_inv_jax(f):
+        return healpix_ifft_jax(f, L, nside, False).real
+
+    def healpix_inv_cuda(f):
+        return healpix_ifft_cuda(f, L, nside, False).real
```
matt-graham, line 162:
```suggestion
    def healpix_inv_jax(ftm):
        return healpix_ifft_jax(ftm, L, nside, False).real

    def healpix_inv_cuda(ftm):
        return healpix_ifft_cuda(ftm, L, nside, False).real
```
For consistency with argument names to `healpix_ifft_*` functions and variables being passed to these functions.

tests/test_healpix_ffts.py
```diff
@@ -92,3 +93,94 @@ def test_healpix_ifft_cuda(flm_generator, nside):
...
+@pytest.mark.skipif(not gpu_available, reason="GPU not available")
+@pytest.mark.parametrize("nside", nside_to_test)
+def test_healpix_ifft_cuda_transforms(flm_generator, nside):
+    L = 2 * nside
+
+    # Generate a random bandlimited signal
+    def generate_flm():
+        flm = flm_generator(L=L, reality=False)
+        flm_hp = samples.flm_2d_to_hp(flm, L)
+        f = hp.sphtfunc.alm2map(flm_hp, nside, lmax=L - 1)
```
matt-graham, line 151: I think we could use `s2fft.inverse(flm, L=L, reality=False, method="jax", sampling="healpix")` here instead of going via `healpy`? Rationale being that I would have a slight preference for minimising the number of additional tests that depend on `healpy` as it we are no longer requiring it as direct dependency for package and in the long run it might be possible to also remove it as a test dependency.

tests/test_healpix_ffts.py
```diff
@@ -92,3 +93,94 @@ def test_healpix_ifft_cuda(flm_generator, nside):
...
+@pytest.mark.skipif(not gpu_available, reason="GPU not available")
+@pytest.mark.parametrize("nside", nside_to_test)
+def test_healpix_fft_cuda_transforms(flm_generator, nside):
+    L = 2 * nside
+
+    # Generate a random bandlimited signal
+    def generate_flm():
+        flm = flm_generator(L=L, reality=False)
+        flm_hp = samples.flm_2d_to_hp(flm, L)
+        f = hp.sphtfunc.alm2map(flm_hp, nside, lmax=L - 1)
```
matt-graham, line 107: I think we could use `s2fft.inverse(flm, L=L, reality=False, method="jax", sampling="healpix")` here instead of going via `healpy`? Rationale being that I would have a slight preference for minimising the number of additional tests that depend on `healpy` as it we are no longer requiring it as direct dependency for package and in the long run it might be possible to also remove it as a test dependency.

s2fft/utils/healpix_ffts.py
```diff
@@ -693,26 +693,30 @@ def ring_phase_shifts_hp_jax(
...
     else:
         raise ValueError(f"fft_type {fft_type} not recognised.")
 
 
-def _healpix_fft_cuda_lowering(ctx, f, *, L, nside, reality, fft_type, norm):
+def _healpix_fft_cuda_lowering(ctx, f, *, L, nside, reality, fft_type, norm, adjoint):
+    assert _s2fft.COMPILED_WITH_CUDA, """
+    S2FFT was compiled without CUDA support. Cuda functions are not supported.
+    Please make sure that nvcc is in your path and $CUDA_HOME is set then reinstall s2fft using pip.
+    """
```
matt-graham, line 718:
```suggestion
    assert _s2fft.COMPILED_WITH_CUDA, """
    S2FFT was compiled without CUDA support. Cuda functions are not supported.
    Please make sure that nvcc is in your path and $CUDA_HOME is set then reinstall s2fft using pip.
    """
```
As [`assert` statements can be disabled when Python is run with optimizations enabled](https://docs.astral.sh/ruff/rules/assert/) I would say better to raise an exception here with the same message. Possibly creating a new custom exception type

```Python
class MissingCUDASupport(Exception):

    def __init__(self):
        super().__init__(
            "S2FFT was compiled without CUDA support. CUDA functions are not supported. "
            "Please make sure that nvcc is in your path and $CUDA_HOME is set "
            "then reinstall s2fft using pip."
        )

**Review by matt-graham — COMMENTED** [review]
> Hello @matt-graham @jasonmcewen @CosmoMatt
> 
> Just a quick PR to wrap up a few stuff
> 
>     1. Updated the binding API to the newest [FFI](https://docs.jax.dev/en/latest/ffi.html)
> 
>     2. Added a vmap implementation of the cuda primitive
> 
>     3. Added a transpose rule which allows jacfwd and jacrev (consequently grad aswell)
> 
>     4. added more tests https://github.com/astro-informatics/s2fft/blob/ASKabalan/tests/test_healpix_ffts.py#L100
> 
>     5. Removed two files which are now no longer needed with the FFI API ([kernel helpers](https://github.com/astro-informatics/s2fft/blob/main/lib/include/kernel_helpers.h)) (so maybe they should be removed from the license section)
> 
>     6. Constrained nanobind to be nanobind >=2.0,<2.6" because of a regression [[BUG]: Regression when using scikit build tools and nanobind wjakob/nanobind#982](https://github.com/wjakob/nanobind/issues/982)
> 
> 
> And finally I added [cudastreamhandler](https://github.com/astro-informatics/s2fft/blob/ASKabalan/lib/include/cudastreamhandler.hpp) which is used to split the XLA provided stream for the VMAP lowering (this header is my own work)
> 
> There is an issue with building [pyssht](https://github.com/astro-informatics/s2fft/actions/runs/14133039257/job/39597964501?pr=290) not sure that this is my fault
> 
> I will check the failing worflows when I get the chance, but in the meantime a review is appreciated

Hi @ASKabalan, sorry for the delay in getting back to you.

This all sounds great - thanks for picking up #237 in particular and for the updates to use the newer FFI interface. 

With regards to the failing workflows - this was probably due to #292 which was fixed in #293. If you merge in latest `main` here that should hopefully resolve the upstream dependency build problems that were causing the test workflows to fail.

I've added some initial review comments below. Will have a closer look next week and try testing this out, but don't have access to GPU machine atm.

**Comment by matt-graham** [comment]
I've tried building, installing and running this on a system with CUDA 12.6 + a NVIDIA A100, and running the HEALPix FFT tests with

```
pytest tests/test_healpix_ffts.py
```

consistently the tests hang when trying to run the first `test_healpix_fft_cuda` instance.

Running just the IFFT tests with 

```
pytest tests/test_healpix_ffts.py::test_healpix_ifft_cuda
```

the tests for both set of test parameters pass.

Trying to dig into this a bit, running the following locally

```Python
import healpy
import jax
import s2fft
import numpy

jax.config.update("jax_enable_x64", True)

seed = 20250416
nside = 4
L = 2 * nside
reality = False

rng = numpy.random.default_rng(seed)
flm = s2fft.utils.signal_generator.generate_flm(rng=rng, L=L, reality=reality)
flm_hp = s2fft.sampling.s2_samples.flm_2d_to_hp(flm, L)
f = healpy.sphtfunc.alm2map(flm_hp, nside, lmax=L - 1)
flm_cuda = s2fft.utils.healpix_ffts.healpix_fft_cuda(f=f, L=L, nside=nside, reality=reality).block_until_ready()
```

raises an error

```
jaxlib.xla_extension.XlaRuntimeError: INTERNAL: CUDA error: : CUDA_ERROR_ILLEGAL_ADDRESS: an illegal memory access was encountered
```

so it looks like there is some memory addressing issue somewhere in the `healpix_fft_cuda` implementation?

ASKabalan:
Thank you

I was able to reproduce with 12.4.1 but not locally with 12.4 

I will take a look

ASKabalan:
@matt-graham I fixed the issue with CUDA 12.4 and above
I cleaned up the code and added docstrings everywhere
There is still a small issue with mac (since the JAX version is not the same)
But I think this is pretty mature to be merged

matt-graham (commented):
Hi @ASKabalan. Thanks for the updates, this is looking great.

I have added some initial comments from a quick high level check.

I also tried running some of our benchmarks via the scripts in the `benchmarks` directory to check everything is running as expected. Results are below (with `method="jax"` corresponding to current native JAX implementation and `method="jax_cuda"` using the CUDA primitive). 

```
forward
(method: jax, L: 64, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: None):
    min(run times):  0.0016s, max(run times):  0.0016s, compile time:     3.9s, peak memory: 1.1e+04B, max(abs(error)):   0.025, floating point ops: 1.2e+06, mem access: 4.3e+06B
(method: jax, L: 128, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: None):
    min(run times):  0.0035s, max(run times):  0.0036s, compile time:     6.9s, peak memory: 1.1e+04B, max(abs(error)):   0.033, floating point ops: 5.1e+06, mem access: 1.9e+07B
(method: jax, L: 256, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: None):
    min(run times):   0.010s, max(run times):   0.010s, compile time:     13.s, peak memory: 1.1e+04B, max(abs(error)):   0.032, floating point ops: 2.2e+07, mem access: 8.2e+07B
(method: jax, L: 512, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: None):
    min(run times):   0.050s, max(run times):   0.050s, compile time:     26.s, peak memory: 1.0e+04B, max(abs(error)):  0.0095, floating point ops: 9.2e+07, mem access: 3.2e+08B
(method: jax_cuda, L: 64, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: None):
    min(run times):  0.0015s, max(run times):  0.0015s, compile time:    0.73s, peak memory: 8.6e+03B, max(abs(error)):   0.036, floating point ops: 5.3e+05, mem access: 3.9e+06B
(method: jax_cuda, L: 128, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: None):
    min(run times):  0.0032s, max(run times):  0.0033s, compile time:    0.87s, peak memory: 8.6e+03B, max(abs(error)):    0.67, floating point ops: 2.1e+06, mem access: 1.5e+07B
(method: jax_cuda, L: 256, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: None):
    min(run times):  0.0096s, max(run times):  0.0096s, compile time:    0.92s, peak memory: 8.6e+03B, max(abs(error)):    0.61, floating point ops: 8.1e+06, mem access: 6.6e+07B
(method: jax_cuda, L: 512, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: None):
    min(run times):   0.049s, max(run times):   0.049s, compile time:     1.3s, peak memory: 8.6e+03B, max(abs(error)):    0.71, floating point ops: 3.1e+07, mem access: 2.5e+08B
inverse
(method: jax, L: 64, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False):
    min(run times):  0.0021s, max(run times):  0.0021s, compile time:     5.3s, peak memory: 8.8e+03B, floating point ops: 1.2e+06, mem access: 4.5e+07B
(method: jax, L: 128, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False):
    min(run times):  0.0046s, max(run times):  0.0046s, compile time:     12.s, peak memory: 8.8e+03B, floating point ops: 5.3e+06, mem access: 3.5e+08B
(method: jax, L: 256, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False):
    min(run times):   0.011s, max(run times):   0.011s, compile time:     27.s, peak memory: 8.8e+03B, floating point ops: 2.2e+07, mem access: 2.7e+09B
(method: jax, L: 512, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False):
    min(run times):   0.056s, max(run times):   0.056s, compile time:     66.s, peak memory: 8.8e+03B, floating point ops: 9.4e+07, mem access: 2.2e+10B
(method: jax_cuda, L: 64, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False):
    min(run times):  0.0017s, max(run times):  0.0017s, compile time:    0.66s, peak memory: 8.6e+03B, floating point ops: 5.1e+05, mem access: 3.7e+06B
(method: jax_cuda, L: 128, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False):
    min(run times):  0.0037s, max(run times):  0.0038s, compile time:    0.67s, peak memory: 8.6e+03B, floating point ops: 2.0e+06, mem access: 1.5e+07B
(method: jax_cuda, L: 256, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False):
    min(run times):  0.0094s, max(run times):  0.0094s, compile time:    0.78s, peak memory: 8.6e+03B, floating point ops: 7.5e+06, mem access: 5.8e+07B
(method: jax_cuda, L: 512, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False):
    min(run times):   0.052s, max(run times):   0.052s, compile time:    0.88s, peak memory: 8.6e+03B, floating point ops: 3.0e+07, mem access: 2.3e+08B
```

In terms of the run and compilation times this all looks great - run times the same or slightly improved for a corresponding bandlimit `L` and compilation times massively reduced (and close to constant in `L` over the range tested). However, there seems to be something slightly odd going on with the round-trip errors (indicated by `max(abs(error))` entries) which are significantly larger for the CUDA primitive version than for the native JAX implementation. For the HEALPix sampling scheme we would expect a non-negligible round-trip error, particularly without iterative refinement, but the size of the errors here seems to be too large and in particular I wouldn't expect there to be a significant difference in the errors compared to the native JAX version. Further when we use iterative refinement with 3 iterations the errors seem to be getting larger when using the CUDA primitive version rather than getting smaller as for the native JAX version.

```
forward
(method: jax_cuda, L: 64, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: 3):
    min(run times):   0.012s, max(run times):   0.012s, compile time:     2.4s, peak memory: 1.1e+04B, max(abs(error)):     49., floating point ops: 3.9e+06, mem access: 2.9e+07B
(method: jax_cuda, L: 128, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: 3):
    min(run times):   0.027s, max(run times):   0.027s, compile time:     2.6s, peak memory: 1.1e+04B, max(abs(error)):     68., floating point ops: 1.5e+07, mem access: 1.2e+08B
(method: jax_cuda, L: 256, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: 3):
    min(run times):   0.074s, max(run times):   0.074s, compile time:     2.7s, peak memory: 1.1e+04B, max(abs(error)): 1.1e+02, floating point ops: 5.5e+07, mem access: 4.9e+08B
(method: jax_cuda, L: 512, L_lower: 0, sampling: healpix, spin: 0, L_to_nside_ratio: 2, reality: True, spmd: False, n_iter: 3):
    min(run times):    0.39s, max(run times):    0.39s, compile time:     3.4s, peak memory: 1.0e+04B, max(abs(error)): 1.7e+02, floating point ops: 2.1e+08, mem access: 1.9e+09B
```

I haven't yet figured out what is causing this. The errors seem to be larger at higher bandlimits which might explain why the tests are not picking this up, but need to investigate this in more detail.

matt-graham:
I have been trying to diagnose what is causing the numerical issues here. Not isolated the precise cause yet, but have somewhat narrowed things down.

- This only seems to affect the forward tranform `s2fft.utils.healpix_ffts.healpix_fft_cuda` and note backward transform `s2fft.utils.healpix_ffts.healpix_ifft_cuda`. Changing `nsides_to_test` in `tests/test_healpix_fft.py` to `nside_to_test = list(range(2, 16)) + [16, 32, 64]` and running `pytest tests/test_healpix_ffts.py::test_healpix_ifft_cuda`, the tests all pass consistently over multiple tries.
- For `s2fft.utils.healpix_ffts.healpix_fft_cuda`, there are localised differences in output compared to `s2fft.utils.healpix_ffts.healpix_fft_jax` for specific `nside` / `L` values, with differences seeming to become more likely for larger `nside`. For example for `nside = 12` we get that indices that the outputs differ in are
   ```
   >>> np.nonzero((abs(ftm_jax - ftm_cuda) > 1e-7))
   (array([10, 10, 10, 10, 10, 10, 10, 10, 21, 21, 21, 21, 21, 21, 21, 21]), 
    array([16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31]))
   ```
- There seems to be some non-determinism in whether there are differences or not when running `pytest tests/test_healpix_ffts.py::test_healpix_fft_cuda` for given `nside` values, despite all input data being the same (the tests used a fixed random seed to generate data). For example for consecutive test runs I got
   ```
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-2] PASSED                                                                                                                                                       [  5%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-3] PASSED                                                                                                                                                       [ 11%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-4] PASSED                                                                                                                                                       [ 17%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-5] FAILED                                                                                                                                                       [ 23%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-6] PASSED                                                                                                                                                       [ 29%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-7] PASSED                                                                                                                                                       [ 35%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-8] PASSED                                                                                                                                                       [ 41%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-9] PASSED                                                                                                                                                       [ 47%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-10] PASSED                                                                                                                                                      [ 52%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-11] PASSED                                                                                                                                                      [ 58%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-12] FAILED                                                                                                                                                      [ 64%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-13] PASSED                                                                                                                                                      [ 70%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-14] FAILED                                                                                                                                                      [ 76%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-15] FAILED                                                                                                                                                      [ 82%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-16] FAILED                                                                                                                                                      [ 88%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-32] FAILED                                                                                                                                                      [ 94%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-64] FAILED                                                                                                                                                      [100%]
   ```
   and then
   ```
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-2] PASSED                                                                                                                                                       [  5%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-3] PASSED                                                                                                                                                       [ 11%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-4] PASSED                                                                                                                                                       [ 17%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-5] PASSED                                                                                                                                                       [ 23%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-6] PASSED                                                                                                                                                       [ 29%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-7] PASSED                                                                                                                                                       [ 35%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-8] PASSED                                                                                                                                                       [ 41%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-9] PASSED                                                                                                                                                       [ 47%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-10] PASSED                                                                                                                                                      [ 52%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-11] PASSED                                                                                                                                                      [ 58%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-12] FAILED                                                                                                                                                      [ 64%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-13] FAILED                                                                                                                                                      [ 70%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-14] PASSED                                                                                                                                                      [ 76%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-15] FAILED                                                                                                                                                      [ 82%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-16] PASSED                                                                                                                                                      [ 88%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-32] FAILED                                                                                                                                                      [ 94%]
  tests/test_healpix_ffts.py::test_healpix_fft_cuda[8966433580120847635-64] FAILED                                                                                                                                                      [100%]
   ```
  with for example `nside=5` failing in the first run and not the second, and `nside=13` failing in the second run but not the first.

The last point in particular makes me think this is something to do with unsafe memory access. I have been looking through the code but haven't spotted anything obvious so far. Unfortunately I also don't seem to be able to get the version before changes in this PR to run - while I can build the CUDA extension module, when running the tests on a GPU the process hangs indefinitely when reaching any calls to the custom primitives.

ASKabalan:
@matt-graham 
 Took some time for to get back at this
 But I think that the race condition is now fixed

matt-graham:
Looking at the code in a bit more detail, I think at least part of what is causing this issue is that it appears that `healpix_forward` in `lib/src/extensions.cc` performs the FFT operations in place on the passed input buffer. Concentrating on the non-batched case for simplicity

https://github.com/astro-informatics/s2fft/blob/3bcb69add7a954b1d552e3f101b576d81891e337/lib/src/extensions.cc#L127-L142

`data_c` is a reinterpreted cast view of the `input` buffer data and is passed to `executor->Forward` along with the workspace buffer view `workspace_c`. The logic of the implementation of `Forward` seems to be to perform the cuFFT operations in place on slices of the `data_c` argument:

https://github.com/astro-informatics/s2fft/blob/3bcb69add7a954b1d552e3f101b576d81891e337/lib/src/s2fft.cu#L130-L131

with pointers to offsets within `data_c` array passed as both `input` and `ouput` arguments to `cufftXtExec`. `data_c` after the call to `Forward` is then passed directly to the `launch_spectral_extension` kernel which writes output to a separate `out_c` output buffer view.

Supporting this hypothesis, if we inspect the values in the mutated underlying `_arrays` property of an array passed to `s2fft.forward` with `method="jax_cuda"`, these do seem to correspond to FFTs of slices of the original input array values corresponding to the HEALPix rings:

```python
import jax
jax.config.update("jax_enable_x64", True)
import s2fft
import numpy as np

from s2fft.sampling import s2_samples

seed = 1234
L = 4
sampling = "healpix"
method = "jax_cuda"
nside = L // 2
reality = True

rng = np.random.default_rng(seed)
kwargs = {"L": L, "reality": reality, "nside": nside, "sampling": sampling}

flm = s2fft.utils.signal_generator.generate_flm(rng=rng, L=L, reality=reality)
f = s2fft.inverse(flm, **kwargs, method="jax")

f_copy = f.copy()

# Required to ensure a cached _npy_value is recorded before mutation
print(f"Before forward call: {np.all(f_copy == f._value) = }")
print(f"Before forward call: {np.all(f_copy == f._arrays[0]) = }")

s2fft.forward(f, **kwargs, method="jax_cuda")

print(f"After forward call: {np.all(f_copy == f._value) = }")
print(f"After forward call: {np.all(f_copy == f._arrays[0]) = }")

index = 0
for t in range(s2_samples.ntheta(L, sampling, nside)):
    nphi = s2_samples.nphi_ring(t, nside)
    fft_slice_matches = np.allclose(
        jax.numpy.fft.fft(f._value[index : index + nphi], norm="backward"),
        f._arrays[0][index : index + nphi]
    )
    print(f"Ring {t}, {fft_slice_matches = }")
    index += nphi
```

outputs

```
Before forward call: np.all(f_copy == f._value) = Array(True, dtype=bool)
Before forward call: np.all(f_copy == f._arrays[0]) = Array(True, dtype=bool)
After forward call: np.all(f_copy == f._value) = Array(True, dtype=bool)
After forward call: np.all(f_copy == f._arrays[0]) = Array(False, dtype=bool)
Ring 0, fft_slice_matches = True
Ring 1, fft_slice_matches = True
Ring 2, fft_slice_matches = True
Ring 3, fft_slice_matches = True
Ring 4, fft_slice_matches = True
Ring 5, fft_slice_matches = True
Ring 6, fft_slice_matches = True
```

There seems to be something more going on that just this however, as if we change the lines

```python
flm = s2fft.utils.signal_generator.generate_flm(rng=rng, L=L, reality=reality)
f = s2fft.inverse(flm, **kwargs, method="jax")
```

to 

```python
f = jax.numpy.asarray(rng.standard_normal(s2_samples.f_shape(L, sampling, nside)))
```

so removing the `s2fft.inverse` call (with `method="jax"`), we instead get output

```
Before forward call: np.all(f_copy == f._value) = Array(True, dtype=bool)
Before forward call: np.all(f_copy == f._arrays[0]) = Array(True, dtype=bool)
After forward call: np.all(f_copy == f._value) = Array(True, dtype=bool)
After forward call: np.all(f_copy == f._arrays[0]) = Array(True, dtype=bool)
Ring 0, fft_slice_matches = False
Ring 1, fft_slice_matches = False
Ring 2, fft_slice_matches = False
Ring 3, fft_slice_matches = False
Ring 4, fft_slice_matches = False
Ring 5, fft_slice_matches = False
Ring 6, fft_slice_matches = False
```

suggesting in this case `s2fft.forward` does not mutate the underlying data in `f` 😕

CMakeLists.txt
```diff
@@ -60,26 +76,35 @@ else()
   if(SKBUILD)
     message(WARNING "CUDA compiler not found, building without CUDA support")
 
-    find_package(Python 3.8 COMPONENTS Interpreter Development.Module REQUIRED)
+    find_package(
+      Python 3.8
```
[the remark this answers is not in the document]
matt-graham: Again think this comment still applies.

PR #296, opened 2025-04-15
Gaussian random coefficients from power spectra

### LINKED ISSUE #201: Compute Gaussian simulation from power spectrum

ntessore:
This PR adds a new function `s2fft.utils.signal_generator.generate_flm_from_spectra()` that generates a stack of Gaussian $f_{lm}$ from a stack of power spectra. Closes #201.

The function is implemented in terms of the existing `generate_flm()`, which gains a new `size=` parameter for generating more than one set of coefficients at a time.

The generated `flm` are then simply scaled by the matrix square root of the provided stack of power spectra. This is implemented here in terms of the SVD, not the Cholesky decomposition, so that it works more reliably with positive semi-definite spectra (e.g. E/B mode spectra for spin-2 fields where the first two entries vanish).

The most difficult part of the PR was the testing: here I am sampling a set of Gaussian random fields, then construct the Gaussian covariance and make sure that the chi2 of the realised fields is within 3 sigma of the expected value. Happy for any suggestions for a better test!

Finally, the whole thing is implemented using the Array API, so that it works with both numpy and jax (and both are tested). For jax, the first-order grads wrt to the spectra are checked as well. There is an issue with the reverse second-order grads that I don't fully understand, so I am not testing those here.

I am marking this as draft for the time being, please share your thoughts on the API and implementation.

### DIFF
--- a/s2fft/utils/signal_generator.py
+++ b/s2fft/utils/signal_generator.py
@@ -8,6 +8,8 @@ from s2fft.sampling import so3_samples as wigner_samples
 
 TYPE_CHECKING = False
 if TYPE_CHECKING:
+    from types import ModuleType
+
     import jax
 
 
@@ -216,6 +218,17 @@ def generate_flmn(
     return torch.from_numpy(flmn) if using_torch else flmn
 
 
+def _get_array_namespace(obj: np.ndarray | jax.Array) -> ModuleType:
+    """Return the correct array namespace for numpy or jax arrays."""
+    from sys import modules
+
+    if (numpy := modules.get("numpy")) and isinstance(obj, numpy.ndarray):
+        return numpy
+    if (jax := modules.get("jax")) and isinstance(obj, jax.Array):
+        return jax.numpy
+    raise TypeError(f"unknown array type: {type(obj)!r}")
+
+
 def generate_flm_from_spectra(
     rng: np.random.Generator,
     spectra: np.ndarray | jax.Array,
@@ -237,8 +250,8 @@ def generate_flm_from_spectra(
         coefficients with the given power spectra.
 
     """
-    # get the Array API namespace from spectra
-    xp = spectra.__array_namespace__()
+    # get an ArrayAPI-ish namespace from spectra
+    xp = _get_array_namespace(spectra)
 
     # check input
     if spectra.ndim != 3 or spectra.shape[0] != spectra.shape[1]:
@@ -262,7 +275,7 @@ def generate_flm_from_spectra(
     a = xp.permute_dims(a, (1, 2, 0))
 
     # sample the random coefficients
-    # always use reality=True, this could be real fields or E/B modes
+    # always use reality=True; one can assemble complex fields from them
     # shape of flm is (K, L, M)
     flm = generate_flm(rng, L, reality=True, size=K)
 
--- a/tests/test_signal_generator.py
+++ b/tests/test_signal_generator.py
@@ -169,7 +169,6 @@ def gaussian_covariance(spectra):
     return cov
 
 
-@pytest.mark.flaky
 @pytest.mark.parametrize("L", L_values_to_test)
 @pytest.mark.parametrize("xp", [np, jnp])
 def test_generate_flm_from_spectra(rng, L, xp):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,6 @@ tests = [
     "numpy<2",  # Required currently due to lack of Numpy v2 compatible pyssht release
     "pytest",
     "pytest-cov",
-    "pytest-rerunfailures",
     "so3",
     "pyssht",
 ]


tests/test_signal_generator.py
```diff
@@ -117,3 +145,93 @@ def test_generate_flmn(rng, L, N, L_lower, reality):
...
+    ell = np.arange(L)
+    for i, j in np.ndindex(row.size, col.size):
+        cov[:, i, j] = (
+            spectra[row[i], row[j]] * spectra[col[i], col[j]]
+            + spectra[row[i], col[j]] * spectra[col[i], row[j]]
+        ) / (2 * ell + 1)
+    return cov
+
+
+@pytest.mark.flaky
```
ntessore, line 170: If `rng` is fixed then this test isn't actually flaky, and `pytest-rerunfailures` isn't needed.

matt-graham: By

> If `rng` is fixed

do you mean if the `numpy.random.Generator` instance returned by the `rng` fixture has a fixed seed? If so, then yes this already the case - there is a default fixed seed set in `tests/conftests.py` and it can be altered by passing a `--seed SEED` argument to `pytest`.

s2fft/utils/signal_generator.py
```diff
@@ -94,29 +99,39 @@ def generate_flm(
...
         np.ndarray: Random set of spherical harmonic coefficients.
 
     """
-    flm = np.zeros(samples.flm_shape(L), dtype=np.complex128)
+    # always turn size into a tuple of int
+    if size is None:
+        size = ()
+    elif isinstance(size, int):
+        size = (size,)
+    elif not (isinstance(size, tuple) and all(isinstance(_, int) for _ in size)):
```
matt-graham, line 113:
```suggestion
    elif not (isinstance(size, tuple) and all(isinstance(i, int) for i in size)):
```
Nitpick, but I think conventionally `_` is used to indicate unused variables in assignments / loops, whereas here the loop variable is being used.

s2fft/utils/signal_generator.py
```diff
@@ -199,3 +214,60 @@ def generate_flmn(
...
+    u, s, vh = xp.linalg.svd(spectra / 2, full_matrices=False)
+
+    # compute the matrix square root for sampling
+    a = u @ (xp.sqrt(s[..., None]) * vh)
+
+    # permute shape (L, K, K) -> (K, K, L)
+    a = xp.permute_dims(a, (1, 2, 0))
+
+    # sample the random coefficients
+    # always use reality=True, this could be real fields or E/B modes
```
matt-graham, line 265:
```suggestion
    # always use reality=True, this could be real fields or E/B modes
```
'E/B modes' should ideally have some clarification here - from a bit of searching it looks like this may be a common term in cosmology settings, but as someone without a cosmology background I don't know what it means and I'd guess I'm representative of this in other potential users and developers from outside cosmology!

tests/test_signal_generator.py
```diff
@@ -76,6 +86,24 @@ def test_generate_flm(rng, L, L_lower, spin, reality):
...
+@pytest.mark.parametrize("L", L_values_to_test)
+@pytest.mark.parametrize("L_lower", L_lower_to_test)
+@pytest.mark.parametrize("spin", spin_to_test)
+@pytest.mark.parametrize("reality", reality_values_to_test)
+def test_generate_flm_size(rng, L, L_lower, spin, reality):
+    if reality and spin != 0:
+        pytest.skip("Reality only valid for scalar fields (spin=0).")
+
+    flm = gen.generate_flm(rng, L, L_lower, spin, reality, size=2)
+    assert flm.shape == (2,) + smp.s2_samples.flm_shape(L)
```
matt-graham, line 98:
```suggestion
    size = 2
    flm = gen.generate_flm(rng, L, L_lower, spin, reality, size=size)
    assert flm.shape == (size,) + smp.s2_samples.flm_shape(L)
```
To avoid magic numbers and make link between function argument and test assertion explicit

tests/test_signal_generator.py
```diff
@@ -76,6 +86,24 @@ def test_generate_flm(rng, L, L_lower, spin, reality):
...
+    check_flm_zeros(flm[0], L, max(L_lower, abs(spin)))
+    check_flm_zeros(flm[1], L, max(L_lower, abs(spin)))
+    check_flm_unequal(flm[0], flm[1], L, max(L_lower, abs(spin)))
+
+    flm = gen.generate_flm(rng, L, L_lower, spin, reality, size=(3, 4))
+    assert flm.shape == (3, 4) + smp.s2_samples.flm_shape(L)
```
matt-graham, line 104:
```suggestion
    size = (3, 4)
    flm = gen.generate_flm(rng, L, L_lower, spin, reality, size=size)
    assert flm.shape == size + smp.s2_samples.flm_shape(L)
```
To avoid magic numbers and make link between function argument and test assertion explicit

s2fft/utils/signal_generator.py
```diff
@@ -199,3 +214,60 @@ def generate_flmn(
...
+
+        spectra (np.ndarray | jax.Array): Stack of angular power spectra.
+
+    Returns:
+        np.ndarray | jax.Array: A stack of random spherical harmonic
+        coefficients with the given power spectra.
+
+    """
+    # get the Array API namespace from spectra
+    xp = spectra.__array_namespace__()
```
matt-graham, line 241: We are not requiring NumPy versions which guarantee Array API support, and this line is therefore causing test failures currently. Would either need to add `array-api-compat` package as a dependency and use that here or just avoid use of Array API for now.

matt-graham (changes requested):
Thanks for this @ntessore I've added some initial review comments below.

> There is an issue with the reverse second-order grads that I don't fully understand, so I am not testing those here.

Out of interest what was is the error / issue you are having here?

s2fft/utils/signal_generator.py
ntessore: I see! If you are only targeting numpy and jax for the time being, then it's probably easiest to write a tiny custom `_get_array_namespace()` function instead of adding an extra dependency for this single function, and ignoring actual Array API compatibility (everything used here exists in both jax and numpy natively).

In any case, I'd rather not code this up as two identical functions that only differ in `np`/`jnp`.

s2fft/utils/signal_generator.py
ntessore: Makes sense! I added a more generic comment now.

In any case, my idea was to only tackle flm with the reality condition here. Sampling flm without that condition is trickier due to the multiple spectra that are required for each field. Even then, the easiest way is to use this "E/B" decomposition into two fields with the reality symmetry, and later assemble the complex fields from there. Ideally, there would be a generic helper function for this conversion (complex <-> E/B), at which point the functionality could be added to `generate_flm_from_spectra()` as well.

matt-graham:
> The auto and numerical derivatives don't match. I suspect this has something to do with the random sampling being called multiple times.


Okay that make sense, I was a bit surprised that closing over the random number generator here by creating an instance with a fixed seed in the outer function and the differentiating through this worked at all, but in retrospect I guess it makes sense as JAX will just see the generated values when it traces the computation and they will then be baked in to the trace as constants.

It does seem possible for this sort of pattern to work with second-order derivatives - for example

```Python
import numpy as np
from jax.test_util import check_grads

def func(x):
    rng = np.random.default_rng(42)
    return x**2 * rng.standard_normal(x.shape)

check_grads(func, (np.array([1., 2.]),), 2)
```

passes without any errors, but I guess this doesn't necessarily carry through to more complex cases.

> The differentiability in light of the random sampling is a bit of a headache. It almost makes more sense to split this functionality up into two parts:
> 
>     1. Sample random coefficients `flm = generate_flm(...)`.
> 
>     2. Scale with the power spectra `flm = apply_power_spectra(spectra, flm)`.
> 
> 
> This would make it straightforward to keep the random coefficients fixed and compute derivatives only wrt. the spectra.

Agree that this sort of design would be better if supporting differentiating through the function is important and it would still also allow exposing a similar interface as currently by having a wrapper function which just calls these two functions.

It also makes it simpler to use the functionality in other contexts - for example a probabilistic programming / Bayesian inference setting. There if you for example wish to sample the random variates using MCMC you would need to pass in values directly into the function. I used a similar approach in [a demo](https://gist.github.com/matt-graham/8ca2fed0a8a23ab32910f825e93af89e) I worked on recently of using `s2fft` to reconstruct a signal on the sphere from noisy samples to allow interfacing with an MCMC package.

s2fft/utils/signal_generator.py
jasonmcewen: Thanks @ntessore , I agree that sounds like a good approach, i.e. considering E/B spectra, since that is the typical convention in cosmology.  Adding some more documentation on E/B fields as @matt-graham suggests is also a good idea.

s2fft/utils/signal_generator.py
jasonmcewen: @matt-graham What approach would you recommend for now?

matt-graham: I think @ntessore's current solution of adding a placeholder `_get_array_namespace` function which just returns the `numpy` or `jax.numpy` namespaces depending on the type of the array argument works well as a solution for now. In the longer term we probably do want to consider using the Array API more generally to avoid some of the duplication across JAX and NumPy implementations of functions Nicolas is trying to avoid here, and I suspect we will want to introduce `array-api-compat` as a dependency at that point to keep NumPy / JAX version requirements loose, but agree that adding this as a dependency for a single function is overkill here!

PR #298, opened 2025-04-16
Some minor updates to #260

matt-graham:
This PR (which is on top of #260) makes some small tidying updates to #260 

- Corrects type hint for argument previously named `DW`
- Changes name of `DW` argument to `precomps` to make use more obvious. I wasn't sure what an appropriate name was here but followed the pattern of using `precomps` as a shorthand for precomputed values in other transform functions.
- Changes internal variable names `Delta`, `Delta_el` and `Quads` to `delta`, `delta_el` and `quads` to make compliant with PEP8 naming conventions.
- Refactors some of the precompute / OTF specific logic in the forward transforms to avoid code repetition.

Tagging you for a review @jasonmcewen, mainly to check you are happy with the change from `DW` to `precomps` as the argument name.

### DIFF
--- a/s2fft/precompute_transforms/fourier_wigner.py
+++ b/s2fft/precompute_transforms/fourier_wigner.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from functools import partial
 
 import jax.numpy as jnp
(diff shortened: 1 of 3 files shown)


jasonmcewen (approved):
Thanks @matt-graham, LGTM.  I think DW was standing for Delta and Weights but precomp seems fine to me since that's what we've typically been using elsewhere (i.e. we haven't been specifying what is stored in the precomp elsewhere).  And the docstring does explain what precomp should contain.

PR #302, opened 2025-04-28
Update README with CUDA extension install details and tweak to Pytorch support note

matt-graham:
Updates notes on HEALPix transform long compilation times with reference to alternatives when running on CPU (JAX `healpy` wrappers) and GPU (CUDA extension module) to sidestep plus adds some basic instructions for how to install with CUDA extension module support. Also tweaks the formatting of the PyTorch support note to remove being in a 'NOTE' alert / callout in favour of just being plain text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@
 </div>
 
 `S2FFT` is a Python package for computing Fourier transforms on the sphere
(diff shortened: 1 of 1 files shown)


jasonmcewen (approved):
Thanks @matt-graham, that's great!

Issue #212, opened 2024-09-16
Segmentation fault on running tests on MacOS on Python 3.9+


matt-graham:
Running `tests/test_spherical_precompute.py::test_transform_inverse_healpix` on a MacOS (`arm64`) runner with Python 3.9 or above generates a segmentation fault [with output](https://github.com/astro-informatics/s2fft/actions/runs/10886545411/job/30206814060)

```
Fatal Python error: Segmentation fault

Thread 0x000000017b2f3000 (most recent call first):
  <no Python frame>

Thread 0x00000002001c8f40 (most recent call first):
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/torch/autograd/graph.py", line 769 in _engine_run_backward
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/torch/autograd/__init__.py", line 436 in grad
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/torch/autograd/gradcheck.py", line 767 in vjp_fn
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/torch/autograd/gradcheck.py", line 894 in _compute_analytical_jacobian_rows
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/torch/autograd/gradcheck.py", line 776 in _check_analytical_jacobian_attributes
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/torch/autograd/gradcheck.py", line 1627 in _slow_gradcheck
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/torch/autograd/gradcheck.py", line 1492 in _gradcheck_real_imag
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/torch/autograd/gradcheck.py", line 2082 in _gradcheck_helper
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/torch/autograd/gradcheck.py", line 2053 in gradcheck
  File "/Users/runner/work/s2fft/s2fft/tests/test_spherical_precompute.py", line 111 in test_transform_inverse_healpix
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/_pytest/python.py", line 159 in pytest_pyfunc_call
[...]

PR #305, opened 2025-05-23
Update `python_requires` and test matrix to support Python 3.11+

matt-graham:
Suggests dropping support for Python 3.8 - 3.10 inline with [SPEC0](https://scientific-python.org/specs/spec-0000/) recommendations, and updating matrix in test workflow to test against Python 3.11, 3.12 and 3.13 on Ubuntu and MacOS. Latter depends on #212 being resolved which may have been achieved by PyTorch refactor in #277.

This also applys various Ruff autofixes for rules that have become active now that we have minimum Python required as 3.11, particularly around use of built in types `list`, `tuple` etc for parameterized type hints, use of `|` rather than `Union` for union types and using `strict` argument to `zip`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,7 +56,7 @@ jobs:
       # by installing torch and healpy using conda
       # https://github.com/healpy/healpy/issues/1012
       run: |
-        conda install jax "jax>=0.3.13,<0.6.0" "numpy>=1.20" ducc0 healpy pytorch pytest pytest-cov
+        conda install jax "jax>=0.3.13,<0.6.0" "numpy>=1.20,<2" ducc0 healpy pytorch pytest pytest-cov
         python -m pip install --upgrade pip
         pip install --no-deps so3 pyssht
         pip install --no-deps .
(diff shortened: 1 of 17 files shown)


jasonmcewen:
Yep, makes sense to drop older python versions

matt-graham:
This is now further updated to use Conda (specifically using `miniforge` installer so using `conda-forge` channel for packages) to install the dependencies on MacOS to avoid the bad interactions between `healpy` and `torch` on MacOS due to incompatible OpenMP runtimes (see https://github.com/astro-informatics/s2fft/issues/212#issuecomment-2956190172). This unfortunately requires us to manually respecify test dependencies in tests workflow file as the test dependencies specified in `pyproject.toml` cannot be used to create a Conda environment. I tried using the in-development [`conda-pypi` plug-in](https://github.com/conda-incubator/conda-pypi) to avoid this, but it appears to not work with current latest Conda version.

As the current `so3` MacOS wheels on PyPI are built with `numpy<2` but there is no NumPy Conda package for Python 3.13 satisfying `1.20<=numpy<2`, I have manually excluded `macos-latest` + Python 3.13 from test matrix currently. Once a new `so3` release is up with wheels built against NumPy v2+ we should be able to remove this exclusion.

PR #315, opened 2025-07-09
Update custom_ops.py

kmulderdas:
Small compatibility change which disables jitting on the s2fft side, in turn enables higher level jitting in s2ai.

### DIFF
--- a/s2fft/precompute_transforms/custom_ops.py
+++ b/s2fft/precompute_transforms/custom_ops.py
@@ -1,8 +1,5 @@
-from functools import partial
-
 import jax.numpy as jnp
 import numpy as np
-from jax import jit
 
 
 def wigner_subset_to_s2(
--- a/s2fft/precompute_transforms/custom_ops.py
+++ b/s2fft/precompute_transforms/custom_ops.py
@@ -86,7 +86,6 @@ def wigner_subset_to_s2(
     return np.fft.ifft(x, axis=-2, norm="forward")
 
 
-# @partial(jit, static_argnums=(3, 4))
 def wigner_subset_to_s2_jax(
     flmn: jnp.ndarray,
     spins: jnp.ndarray,
@@ -209,7 +208,6 @@ def so3_to_wigner_subset(
     return s2_to_wigner_subset(x, spins, DW, L, sampling)
 
 
-# @partial(jit, static_argnums=(3, 4, 5))
 def so3_to_wigner_subset_jax(
     f: jnp.ndarray,
     spins: jnp.ndarray,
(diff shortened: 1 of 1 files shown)


s2fft/precompute_transforms/custom_ops.py
```diff
@@ -86,7 +86,7 @@ def wigner_subset_to_s2(
     return np.fft.ifft(x, axis=-2, norm="forward")
 
 
-@partial(jit, static_argnums=(3, 4))
+# @partial(jit, static_argnums=(3, 4))
```
matt-graham, line 89:
```suggestion
```

We generally shouldn't comment out code as we can always recover snippets from git history - this is likely to be what is causing the linting failures.

s2fft/precompute_transforms/custom_ops.py
```diff
@@ -209,7 +209,7 @@ def so3_to_wigner_subset(
     return s2_to_wigner_subset(x, spins, DW, L, sampling)
 
 
-@partial(jit, static_argnums=(3, 4, 5))
+# @partial(jit, static_argnums=(3, 4, 5))
```
matt-graham, line 212:
```suggestion
```

s2fft/precompute_transforms/custom_ops.py
```diff
@@ -338,7 +338,7 @@ def s2_to_wigner_subset(
     return x * (2.0 * np.pi) ** 2
 
 
-@partial(jit, static_argnums=(3, 4))
+# @partial(jit, static_argnums=(3, 4))
```
matt-graham, line 341:
```suggestion
```

matt-graham (changes requested):
@kmulderdas - is having these functions jitted blocking you jitting a function which calls them in `s2ai`? Generally `jax.jit` wrapped functions [should be able to be arbitrarily nested](https://github.com/jax-ml/jax/discussions/10104) and there is some suggestion it can help with compilation times when a lower-level function is reused multiple times in a higher level function. I suspect removing the `jit` decorators on the utility functions here won't have any major performance implications but it would be good to understand why its causing issues as we widely apply `jax.jit` to other functions in `s2fft` which are likely to be contained in higher-level jitted functions.

s2fft/precompute_transforms/custom_ops.py
kmulderdas: These jitted functions are called in the lifted convolution layers in `s2ai`. For performance and general functionality reasons it is desirable to have the flax model, build from these layers, be traceable and jittable at the top level. These utility functions specifically cause errors when trying to trace/jit at the aforementioned level. It is not clear to me why the other imported jitted functions from `s2fft` or the ones natively defined in `s2ai` don't break in the same way.

matt-graham (approved):

PR #351, opened 2026-01-20
Documentation for sampling methods

### LINKED ISSUE #306: Document details of sampling schemes provided in package
At the moment while we have some documentation about the [spherical sampling schemes supported in the main README](https://github.com/astro-informatics/s2fft?tab=readme-ov-file#sampling-) this does not give details of the differences between the sampling schemes and when one might be favoured over another, nor does it explicitly indicate how these are specified in our API.

willGraham01:
Resolves #306 |

Adds a new page to the documentation providing some details and references for the sampling schemes supported by `s2fft`. Also adds some text on how these are specified in the package API, a mapping of said string specifications to the schemes themselves, and a short example of providing the `sampling` keyword to some of the methods within the package.

Some things that would be worth reviewing:

- The current "at a glance" table should be checked for correctness (rst is a nightmare so I don't doubt that I've got a column or two wrong somewhere). Also, if there are any other headings + columns that we want to add to this quick comparison table, then shout below.
- The example to demonstrate how to use the `sampling` keyword (we might be able to think up a better example)
- Where we want this info. I've added it as a new page currently but it could be chopped up + moved somewhere else.


docs/sampling.rst
```diff
@@ -0,0 +1,201 @@
...
+  from s2fft.base_transforms.spherical import forward, inverse
+  from s2fft.sampling.s2_samples import thetas, phis_equiang
+
+  L = 512
+
+  def signal(theta, phi):
+      """Our known signal function, that we will 'sample' from."""
+      return jnp.sin(theta)**2 * jnp.sin(phi)
+
+We can generate arrays containing the :math:`\theta_t` and :math:`\varphi_p` sample coordinates for the MW and GL schemes using :func:`~s2fft.sampling.s2_sampling.thetas` and :func:`~s2fft.sampling.s2_sampling.phis_equiang`.
```
matt-graham:
```suggestion
We can generate arrays containing the :math:`\theta_t` and :math:`\varphi_p` sample coordinates for the MW and GL schemes using :func:`~s2fft.sampling.s2_samples.thetas` and :func:`~s2fft.sampling.s2_samples.phis_equiang`.
```

docs/sampling.rst
```diff
@@ -0,0 +1,201 @@
+Sampling
+========
+
+.. image:: https://raw.githubusercontent.com/astro-informatics/s2fft/main/docs/assets/figures/spherical_sampling.png
+    :width: 700
+    :alt: Visualization of spherical sampling schemes
+    :align: center
```
matt-graham:
```suggestion
    :align: center
    :class: dark-light
```

This will remove the white background when docs are viewed in dark mode.

docs/sampling.rst
```diff
@@ -0,0 +1,201 @@
...
+  # Generate (theta, phi) points used by the GL scheme
+  gl_thetas, gl_phis = thetas(L, sampling="gl"), phis_equiang(L, sampling="gl")
+
+  # We now pretend we have two different observations of this signal,
+  # but which sampled it using different schemes.
+  mw_signal_samples = signal(*jnp.meshgrid(mw_thetas, mw_phis, indexing='ij'))
+  gl_signal_samples = signal(*jnp.meshgrid(gl_thetas, gl_phis, indexing='ij'))
+
+Now that we have two sets of samples from the same signal, we can forward transform obtain the harmonic coefficients.
+In each case, we need to specify which sampling scheme was used to obtain the data, by passing the ``sampling`` argument to the :func:`~s2fft.base_transforms.spherical.forward` transform when we call it.
```
matt-graham:
```suggestion
In each case, we need to specify which sampling scheme was used to obtain the data, by passing the ``sampling`` argument to the :func:`~s2fft.transforms.spherical.forward` transform when we call it.
```

docs/sampling.rst
```diff
@@ -0,0 +1,201 @@
...
+      - No
+      - Yes
+    * - :ref:`driscoll-healy-dh`
+      - ``"dh"``
+      - Yes
+      - No
+      - Yes
+    * - :ref:`guass-legendre-gl`
+      - ``"gl"``
+      - Yes
```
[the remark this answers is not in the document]
willGraham01: (this is why you never copy-paste columns when creating an rST table :sweat_smile: )

docs/sampling.rst
```diff
@@ -0,0 +1,201 @@
...
+It also requires fewer sampling points than the GL scheme, though asymptotically the number of sampling points used by GL is the same as MW.
+
+Sample positions are defined by
+
+.. math::
+
+  \theta_t  &= \frac{\pi (2t+1)}{2L-1}, &\quad t\in\lbrace 0,1,...,L-1 \rbrace, \\
+  \varphi_p &= \frac{2\pi p}{2L-1},     &\quad p\in\lbrace 0,1,...,2L-2\rbrace.
+
+The total number of samples is :math:`N_{MW} = (L-1)(2L-1)+1 \sim 2L^2`.
```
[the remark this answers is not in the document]
willGraham01: All samples with $\theta = \pi$ are at the same point though (achieved when $t = L-1$)?

- ($\pi, \phi_{p}$) is the same point for every $\phi_{p}$ in the grid, so we only get 1 sample site from here.
- We then have $L - 1$ remaining $\theta$ values, each of which as $2L- 1$ $\phi$ values, giving $(L-1)(2L-1)$ more sample points.

PR #362, opened 2026-02-20
Include Windows in test and build workflows and allow package use without extension module

### LINKED ISSUE #360: DLL issue on import
Hello! I have been trying to import the package, but whenever I do so, I get the following error:

```
[...]
 File "C:\Users\kam224\AppData\Local\Python\pythoncore-3.14-64\Lib\site-packages\s2fft\utils\__init__.py", line 1, in <module>
    from . import (
    ...<11 lines>...
    )
 File "C:\Users\...\AppData\Local\Python\pythoncore-3.14-64\Lib\site-packages\s2fft\utils\healpix_ffts.py", line 11, in <module>
    from s2fft_lib import _s2fft
ImportError: DLL load failed while importing _s2fft: The specified module could not be found.
```

The issue has persisted despite clean python installs with versions 3.11 and 3.9. I have tried installing using pip by name and by github link. The only dll missing when viewing the pyd file using the Dependencies program, seemingly, looks to be python314.dll. But despite adding the dll's folder location using the os package, or moving the dll to the same folder, the issue persists.

Would anyone have any ideas on how to fix this?

matt-graham:
Potentially resolves #360 

- Makes import of `s2fft_lib` extension module optional by wrapping in `try: ... except ImportError: ...` logic as this should allow package to still be used (without functionality in extension module) when the extension module cannot be built and a pre-built binary is not available.
- Adds Windows runners to (cached test data) tests matrix. Previously this was blocked by lack of Windows support in some our test dependencies like `ssht` and `so3` but as these are only required for the tests which produced cached data, we may be able to run tests with cached data on Windows now.

Keeping as draft for now as I assume there are other things that will be needed to get Windows support working that I've missed.

### DIFF
--- a/s2fft/utils/healpix_ffts.py
+++ b/s2fft/utils/healpix_ffts.py
@@ -1,4 +1,5 @@
 from functools import partial
+from warnings import warn
 
 import jax
 import jax.numpy as jnp
@@ -16,6 +17,7 @@ except ImportError:
     # efficient HEALPix FFTs. As extension module may not compile on all
     # systems we guard import here to allow still using the package when the
     # module is not available
+    warn("s2fft_lib extension module not available", stacklevel=-2)
     _s2fft = None
 
 from s2fft.sampling import s2_samples as samples


s2fft/utils/healpix_ffts.py
```diff
@@ -806,9 +831,10 @@ def _healpix_fft_cuda_transpose(
...
-# Register healpfix_fft_cuda custom call target
-for name, fn in _s2fft.registration().items():
-    jax.ffi.register_ffi_target(name, fn, platform="CUDA")
+if _s2fft is not None:
+    # Register healpfix_fft_cuda custom call target
+    for name, fn in _s2fft.registration().items():
+        jax.ffi.register_ffi_target(name, fn, platform="CUDA")
 
 # Step 1: Register the HEALPix FFT CUDA primitive with JAX.
 _healpix_fft_cuda_primitive = register_primitive(
```
willGraham01: I found it a bit odd to have conditional statements at module-level sandwiched between function definitions halfway through the file - initially thought that they were meant to be part of another function 😅 Maybe this chunk could move to line 933 (after the last function is defined?).

Also, I don't think the "Step 1:" part of the register comment is necessary... there are no subsequent steps marked by comments?

tests/test_healpix_ffts.py
```diff
@@ -214,3 +218,15 @@ def healpix_inv_cuda(ftm):
...
+def test_extension_module_not_available_raise_error(monkeypatch):
+    monkeypatch.setattr(sys.modules["s2fft.utils.healpix_ffts"], "_s2fft", None)
+    with pytest.raises(MissingExtensionModule):
+        _check_extension_module()
+
+
+def test_extension_module_no_cuda_support_raise_error(monkeypatch):
+    monkeypatch.setattr(sys.modules["s2fft_lib._s2fft"], "COMPILED_WITH_CUDA", False)
+    with pytest.raises(MissingCUDASupport):
+        _check_extension_module()
```
willGraham01:
```suggestion
def test_extension_module_not_available_raise_error(monkeypatch):
    monkeypatch.setattr(sys.modules["s2fft.utils.healpix_ffts"], "_s2fft", None)
    with pytest.raises(
        MissingExtensionModule, match="The s2fft_lib extension module is not available."
    ):
        _check_extension_module()


def test_extension_module_no_cuda_support_raise_error(monkeypatch):
    monkeypatch.setattr(sys.modules["s2fft_lib._s2fft"], "COMPILED_WITH_CUDA", False)
    with pytest.raises(
        MissingCUDASupport, match="S2FFT was compiled without CUDA support."
    ):
        _check_extension_module()
```

Possibly overkill as the exception subclasses always init with the same message, but it's better to make sure in case we do want to generalise the error messages in future.

PR #380, opened 2026-05-20
Replace custom_vjp with registered JAX primitives for latitudinal step

### LINKED ISSUE #379: Replacing custom_vjp on latitudinal step with custom JAX primitives (custom_vjp lacks a transpose)
The latitudinal Wigner-d recursion in s2fft/transforms/spherical.py is wrapped in jax.custom_vjp. This works for jax.grad, but custom_vjp does not declare a transpose rule, so jax.linear_transpose cannot peel through the latitudinal step — even though the step is linear and has an exact analytical transpose (the forward step is the transpose of the inverse step and vice versa).

I tried to fix it with jax.experimental.custom_transpose,but unfortunately I couldn't find this API (using JAX 0.10.0). 

Instead I propose solving it by registering the forward and inverse latitudinal steps as custom JAX primitives similar to ffts in s2fft/utils/healpix_ffts.py with an explicit transpose rule (plus a JVP rule and a batcher), instead of custom_vjp. I have my solution to this issue on my fork. 

**The idea**

jax.custom_vjp declares a VJP — fwd / bwd — but does not declare how to transpose the computation as a linear map. Any caller that goes through jax.linear_transpose (directly, or indirectly via JAX transforms that lower to it) cannot transpose the latitudinal step analytically and will fail. For example, when calling the transpose of s2fft.transforms.spherical.inverse_jax.

The operation is linear, so jax.linear_transform should be the API to call, however it fails without custom transpose rule declared. 

**Proposed solution**

Register two custom JAX primitives — flm_to_ftm (inverse latitudinal step) and ftm_to_flm (forward latitudinal step) — using the existing s2fft.utils.jax_primitive.register_primitive helper that's already used for the HEALPix CUDA targets.

Each primitive declares:

- abstract eval — output shape/dtype from samples.ftm_shape samples.flm_shape,
- MLIR lowering — mlir.lower_fun of the existing otf.*_latitudinal_step_jax (numerics unchanged),
- JVP rule — linear ⇒ tangent is the primitive re-bound on the tangent input,
- explicit transpose rule — the actual point of this change: the transpose of flm_to_ftm is
ftm_to_flm, and vice versa,
- batching rule — lifts every operand to a shared leading batch dim and re-binds the primitive, so the analytical-transpose path survives vmap.

Sketch of the transpose rule, which is the part that solves jax.linear_transpose:

```
def _flm_to_ftm_transpose(cotangent, flm, thetas, spin, *precomps,
                          L, nside, sampling, reality, spmd, L_lower):
    # Analytical adjoint: transpose of the inverse step IS the forward step.
    # `precomps` are direction-specific, so pass None to regenerate the
    # forward-direction coefficients.
    cot_flm = _underlying_forward(
        cotangent, thetas, spin, None,
        L=L, nside=nside, sampling=sampling,
        reality=reality, spmd=spmd, L_lower=L_lower,
    )
    return (cot_flm, None, None) + (None,) * len(precomps)

_flm_to_ftm_primitive = register_primitive(
    "flm_to_ftm",
    multiple_results=False,
    abstract_evaluation=_flm_to_ftm_abstract,
    lowering_per_platform={None: mlir.lower_fun(_flm_to_ftm_impl,
                                                multiple_results=False)},
    batcher=_flm_to_ftm_batcher,
    jacobian_vector_product=_flm_to_ftm_jvp,
    transpose=_flm_to_ftm_transpose,
    is_linear=False,  # marked False because spin/precomps are operands
)
```

Call sites in spherical.py collapse to a single function call replacing the custom_vjp

```
ftm = ftm_flm_prim.flm_to_ftm(
    flm, thetas, L=L, spin=spin, nside=nside, sampling=sampling,
    reality=reality, spmd=spmd, L_lower=L_lower, precomps=precomps,
)
```

With this in place, jax.linear_transpose succeeds and routes through the analytical adjoint instead of failing.

ArtemBasyrov:
## Summary
  
  Closes #379.

The latitudinal Wigner-d step in `spherical.inverse_jax` / `spherical.forward_jax` was previously wrapped with `jax.custom_vjp`. While that is sufficient for `jax.grad`, `custom_vjp` does not declare a transpose rule, so `jax.linear_transpose` (and any transform that relies on it — e.g. `jax.linearize`, double-`vjp`, certain `vmap` + autodiff compositions) fails on these transforms even though the operation is linear and has an exact analytical adjoint.

This PR replaces the `custom_vjp` wrapping with two registered JAX primitives, `flm_to_ftm` and `ftm_to_flm`, defined in a new module `s2fft/transforms/_ftm_flm_primitive.py`. Each primitive declares:
  
  - an **abstract eval** for shape/dtype propagation,
  - an **MLIR lowering** that preserves the existing numerics,
  - a **JVP rule** for tangent propagation,
  - an explicit **transpose rule** — the key fix — where the transpose of `flm_to_ftm` is `ftm_to_flm` and vice-versa, exploiting the analytical adjoint,
  - a **batching rule** so `vmap` composes cleanly with autodiff.

The two primitives are wired through the existing `register_primitive` helper already used for the HEALPix CUDA targets, so the registration path is consistent with the rest of the codebase.
  
## Changes

  - `s2fft/transforms/_ftm_flm_primitive.py` *(new)* — primitive definitions, lowerings, JVP / transpose / batching rules.
  - `s2fft/transforms/spherical.py` — `inverse_jax` and `forward_jax` now call the primitives instead of the `custom_vjp`-wrapped helpers; dead `custom_vjp` plumbing removed.
  - `tests/test_ftm_flm_primitive.py` *(new)* — tests covering forward numerics, `jvp`, `vjp`, `linear_transpose`, double-transpose round-trip, and `vmap` interactions.

### DIFF
--- a/s2fft/transforms/_ftm_flm_primitive.py
+++ b/s2fft/transforms/_ftm_flm_primitive.py
@@ -167,9 +167,7 @@ def _lift_to_batch(arr, ax, batch_size):
     return jnp.moveaxis(arr, ax, 0)
 
 
-def _flm_to_ftm_batcher(
-    batched_args, batch_axes, *, L, nside, sampling, reality, spmd, L_lower
-):
+def _flm_to_ftm_batcher(batched_args, batch_axes, **params):
     flm, thetas, spin, *precomps = batched_args
     flm_ax, thetas_ax, spin_ax, *precomps_ax = batch_axes
     if thetas_ax is not None:
@@ -177,14 +175,6 @@ def _flm_to_ftm_batcher(
             "vmap over `thetas` is not supported (it is determined by the "
             "static sampling configuration)."
         )
-    params = dict(
-        L=L,
-        nside=nside,
-        sampling=sampling,
-        reality=reality,
-        spmd=spmd,
-        L_lower=L_lower,
-    )
     # Identify batch size from the first batched operand.
     arrays_axes = [(flm, flm_ax), (spin, spin_ax)] + list(
         zip(precomps, precomps_ax, strict=False)
@@ -306,9 +296,7 @@ def _ftm_to_flm_transpose(
     return (cot_ftm, None, None) + (None,) * len(precomps)
 
 
-def _ftm_to_flm_batcher(
-    batched_args, batch_axes, *, L, nside, sampling, reality, spmd, L_lower
-):
+def _ftm_to_flm_batcher(batched_args, batch_axes, **params):
     ftm, thetas, spin, *precomps = batched_args
     ftm_ax, thetas_ax, spin_ax, *precomps_ax = batch_axes
     if thetas_ax is not None:
@@ -316,14 +304,6 @@ def _ftm_to_flm_batcher(
             "vmap over `thetas` is not supported (it is determined by the "
             "static sampling configuration)."
         )
-    params = dict(
-        L=L,
-        nside=nside,
-        sampling=sampling,
-        reality=reality,
-        spmd=spmd,
-        L_lower=L_lower,
-    )
     arrays_axes = [(ftm, ftm_ax), (spin, spin_ax)] + list(
         zip(precomps, precomps_ax, strict=False)
     )
--- a/s2fft/transforms/_ftm_flm_primitive.py
+++ b/s2fft/transforms/_ftm_flm_primitive.py
@@ -47,40 +47,6 @@ from s2fft.utils.jax_primitive import register_primitive
 _DATA_NDIM = 2
 
 
-def _underlying_inverse(
-    flm, thetas, spin, precomps, *, L, nside, sampling, reality, spmd, L_lower
-):
-    return otf.inverse_latitudinal_step_jax(
-        flm,
-        thetas,
-        L,
-        spin,
-        nside,
-        sampling,
-        reality,
-        precomps=precomps,
-        spmd=spmd,
-        L_lower=L_lower,
-    )
-
-
-def _underlying_forward(
-    ftm, thetas, spin, precomps, *, L, nside, sampling, reality, spmd, L_lower
-):
-    return otf.forward_latitudinal_step_jax(
-        ftm,
-        thetas,
-        L,
-        spin,
-        nside,
-        sampling,
-        reality,
-        precomps=precomps,
-        spmd=spmd,
-        L_lower=L_lower,
-    )
-
-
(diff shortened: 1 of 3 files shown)


matt-graham:
Thanks @ArtemBasyrov for working on this - I'll try get to reviewing this next week. From high-level what you describes sounds sensible and something we'd been vaguely aware we should do to allow higher-order automatic differentiation / composing with batching, so it's great that you've done this!

matt-graham:
Hi @ArtemBasyrov. I've now done a review through the changes here and functionally seem generally great, but I have opened up a PR on top of this (https://github.com/ArtemBasyrov/s2fft/pull/1) which suggests some changes which I think improve the readability of the code (in terms of making the logic a bit clearer by reducing the boiler plate associated with passing around parameters) and also make the gradient tests a bit stricter (by testing both forward and reverse mode and with default tolerances). I have also suggested removing the end-to-end tests added here as while a good idea to check during development, these are redundant given the existing tests in `tests/test_spherical_transforms.py` which check the forward and inverse transforms across different methods (including JAX and NumPy implementations) are consistent with reference values computed with external packages (and so by implication are consistent with each other).

My suggested changes to use keyword parameter syntax / unpacking in the primitive functions do end up making these a bit less robust than your approach of explicitly specifying all keyword arguments for each function as we would silently pass through invalid keyword argument names, but as this is only done here for the internal functions not the public functions, I think the slight loss of robustness is probably worth the code reduction.

I also noticed during my updates to tests that we currently get incorrect gradients when evaluating second-order derivatives using reverse-mode with MWSS sampling scheme for `ftm_to_flm` primitive, which is why I have restricted the check to `order=1` in that case for now. From what I can tell so far this is not related to your changes and I think may relate to handling of the pole singularities for MWSS and as I think we didn't previously support higher-order autodiff so this is not breaking existing functionality, I don't think this is a blocker.

If you are happy with my changes, if you merge them in to your fork branch that should trigger a run of CI here. Apologies for the slightly odd workflow - I originally started using the GitHub PR review suggest changes but as my changes ended up touching quite a lot of bits of code it was difficult to keep this all in sync and ensure I wasn't breaking things in turn.

ArtemBasyrov:
Thank you for going through it @matt-graham , your commits did improve readability quite a lot. I have now merged it into my fork.

matt-graham (approved):
Let's get this merged!

matt-graham:
@all-contributors please add @ArtemBasyrov  for code, test

Issue #340, opened 2026-01-06
Spherical grid modularization


matt-graham:
[S2FFT currently supports several spherical grids](https://astro-informatics.github.io/s2fft/#sampling) which define the points on a sphere at which a signal is sampled along with a quadrature rule for computing integrals. The grids supported by S2FFT are biased towards those common in space science. While the supported grids overlap with those common in fields such as climate science, some vary in particular details in the climate setting (for example use of [Clenshaw-Curtis quadrature in equiangular schemes](https://doi.org/10.1002/qj.3282)) while others such as the [octahedral reduced grid used by ECMWF ](https://www.ecmwf.int/sites/default/files/elibrary/2016/17262-new-grid-ifs.pdf) are not supported at all currently.

Multiple grid support is implemented in the package by branching on an argument specifying the scheme to use. The high degree of branching reduces code readability, and makes it difficult to support additional grids as changes need to be applied to multiple different points in the codebase.

An alternative would be to refactor the grid support to use an object-oriented design, with a common interface defined for evaluating the grid specific parts of the transform computations. This would reduce code complexity and make it straightforward to support additional schemes which are variations of those currently supported by subclassing. This would simplify use of the package in climate applications without needing to manually adjust for differences in grid conventions.

PR #381, opened 2026-05-21
Add support for Clenshaw-Curtis and Fejér second rule quadrature / sampling schemes

### LINKED ISSUE #251: [Feature request] Cylindrical sampling scheme
In CMB data analysis (aiming at probing small angular scales), an interest has been growing towards cylindrical griding, named CAR in https://phy-act1.princeton.edu/public/saiola/act_dr4_C20.pdf or ECP in https://www.aanda.org/articles/aa/full_html/2013/08/aa21575-13/aa21575-13.html. It corresponds to a rectangular projection, equi-spaced in latitude, and not equal-area projection, but with the same number of pixels per ring, and relies on Clenshaw-Curtis quadrature weightings for the associated transforms.

This sampling scheme is the baseline one used for the ACT (Atacama Cosmology Telescope) collaboration and central in many tools developed for the LAT analysis in the Simons Observatory collaboration (in which I am part of).

I wanted to ask if it would be possible to have s2fft support for such sampling scheme.

matt-graham:
Should resolve #251 and also does some work towards #340

Adds implementations of the Clenshaw-Curtis and Fejér's (second) quadrature rules and corresponding sampling schemes.

The quadrature weights are computed using the FFT based approach described in 

> Waldvogel, J. (2006). Fast construction of the Fejer and Clenshaw-Curtis quadrature rules.
> BIT Numerical Mathematics, 46(1), 195–202. https://doi.org/10.1007/s10543-006-0045-4.

Following the definitions in that paper the Clenshaw-Curtis scheme includes samples at both poles ($\theta = \pm \pi$) while the Fejér's second rule scheme excludes both poles. The Clenshaw-Curtis scheme is selected with `sampling="cc"` and Fejér's second rule with `sampling="f2"` arguments to relevant functions. 

To facilitate adding these new schemes, this PR also does some light refactoring of the existing quadrature and sample related modules to try to make the code more DRY and with a more standardized interfaced across the different schemes. In some cases the rationale for special casing code paths on different sampling schemes was a bit unclear, particularly around handling of pole singularities and setting `m_offset`,  so I've done a best efforts attempt at trying to rationalise this but this probably could do with some careful checking.

Currently for a bandlimit $L$ both schemes have $N_\theta = 2L - 1$ and $N_\phi = 2L$. The former appears to be required to be $\sim 2L$ to give round-trip errors close to machine precision; I believe this related to the comment

> An important difference [of Clenshaw-Curtis and Fejér rules] from Gaussian quadrature is that the number of nodes, given the truncation total wavenumber, $N$ required for alias-free exact meridional integration, is 
> $J \geq 2N + 1$, 
> which is about twice that in the Gaussian case. This is because Clenshaw–Curtis quadrature expands the integrand into Chebyshev polynomials of up to $(J −1)\text{th}$ degree and thus $J − 1$ needs to be no less than $2N$, the maximum degree of the integrand polynomial.

in [the paper _Hotte and Ujiie (2018) A nestable, multigrid-friendly grid on a sphere for global spectral models based on Clenshaw–Curtis quadrature_](https://doi.org/10.1002%2Fqj.3282) with their definition of $N$ corresponding to $L  - 1$ and $J$ to $N_\theta$ in our notation and hence we require $N_\theta \geq 2 L - 1$. 

Choosing $N_\theta = 2L - 1$  gives an odd number of nodes in the latitude axis and so includes the equator as a node, this matching the co-latitude points used in Hotte and Ujiie (2018). This still leaves $N_\phi$ undetermined - here I have chosen $N_\phi = 2L$ which is sufficient to avoid aliasing in FFT operations (which requires $N_\phi > 2L - 2$), and matches the MW scheme, but gives $N_\theta \approx N_\phi$ while it seems quite common in grids used in practice in climate applications to have $N_\phi \approx 2N_\theta$ which might suggest using $N_\phi = 4L$.

@jasonmcewen I'm tagging you as you mentioned you would be interested in looking through the details of this.

TODO:

- [x] Decide if we want to adjust spherical grid dimensions for given bandlimit $L$ for new schemes
- [x] Add details of new schemes to sampling scheme documentation overview page
- [ ] Check if new schemes can be carried through to Wigner transforms + document


s2fft/precompute_transforms/construct.py
```diff
@@ -13,6 +13,32 @@
...
+    if sampling == "mw":
+        return 2 * n_theta - 1
+    elif sampling == "mwss":
+        return 2 * n_theta - 2
+    elif sampling == "dh":
+        return 2 * n_theta
+    elif sampling == "cc":
+        return 2 * n_theta - 2
+    elif sampling == "f2":
+        return 2 * n_theta + 2
```
matt-graham, line 37: @jasonmcewen I arrived at these by trial and error as I could not see the pattern in how the values for the other schemes was derived - if there some underlying relationship here it would be good to document. It might also be worth moving this to one of modules under `s2fft.sampling`

s2fft/sampling/s2_samples.py
```diff
@@ -1,5 +1,10 @@
 import numpy as np
 
+M_OFFSET_1_SCHEMES = frozenset(("mwss", "healpix", "cc", "f2"))
```
matt-graham: I am not sure if there is a more descriptive / explanatory name for what unites these schemes. I found that including `cc` and `f2` in this set was required to get tests passing but this more by trial and error than due to having a strong handle on what property of schemes requires us to offset `m` indices like this.

s2fft/utils/quadrature.py
```diff
@@ -18,7 +21,7 @@ def quad_weights_transform(
         L (int): Harmonic band-limit.
 
         sampling (str, optional): Sampling scheme.  Supported sampling schemes include
-            {"mwss", "dh", "gl", "healpix}.  Defaults to "mwss".
+            {"mwss", "dh", "gl", "healpix", "cc"}.  Defaults to "mwss".
```
matt-graham: Add `"f2"` to list here

tests/test_quadrature.py
```diff
@@ -48,3 +50,80 @@ def test_quadrature_exceptions():
...
+    "f_and_integral",
+    [
+        (lambda x: x, 0.0),
+        (lambda x: x**2, 2 / 3),
+        (lambda x: x**3 - 2 * x**2 + x - 1, -10 / 3),
+        (lambda x: x**4 - x**2, -4 / 15),
+    ],
+)
+@pytest.mark.parametrize("rule", ["cc", "f2", "mw", "mwss", "gl", "dh"])
+@pytest.mark.parametrize("n_points", [6, 8, 16])
```
matt-graham: Check whether minimum number of points here corresponds to critical threshold

s2fft/sampling/s2_samples.py
matt-graham: Related to whether number of $\phi$ samples is even / oversampling in $\phi$.

matt-graham:
Copying discussion from Slack here so we have a record going forwards:

The Hotte and Ujiie (2018) paper suggests we require `n_theta >= 2 * L - 1` to get round trip errors close to machine precision, which underlies the choice in my current PR to use `n_theta = 2 * L - 1` (along with `n_phi = 2 * L` for the Clenshaw-Curtis `sampling="cc"` and Fejer second rule `sampling="f2"` sampling schemes.

The [`analysis_2d` function in `ducc0`](https://gitlab.mpcdf.mpg.de/mtr/ducc/-/blob/6c871af7d26fd549a94f92f4a9d53b68245b996e/src/ducc0/sht/sht.cc#L1747-1753) however only requires that `n_theta >= nrings_min = lmax + 2 = L + 1` for the Clenshaw-Curtis `geometry="CC"` scheme (and `n_theta >= nrings_min = 2 * lmax + 1 = 2 * L - 1` for the Fejer second rule `geometry="F2"` scheme), with `L = lmax + 1`.



Generating a set of random coefficients then running `ducc0.sht.synthesis_2d` → `ducc0.sht.analysis_2d` to generate map then recover coefficients, using `geometry="CC"` with `ntheta = L + 1`  we do get round-trip errors close to machine precision $\sim 10^{-15}$. Running our corresponding S2FFT implementations `s2fft.inverse` → `s2fft.forward` with `sampling="cc"` and `n_theta = L + 1`  (enforced by manually editing `n_theta` function in source code as we don't allow passing this as an argument unlike `ducc0`) we however get $\sim 1$ round trip errors. The outputs of `ducc0.sht.synthesis_2d` and `s2fft.inverse` for equivalent input coefficients do however match to within machine precision with `sampling="cc"` which confirms as expected this is to do with the implementation of the quadrature specifically.

Looking at the implementation of `ducc0.sht.analysis_2d` specifically [the lines 1763-1791 in commit `6c871af`](https://gitlab.mpcdf.mpg.de/mtr/ducc/-/blob/6c871af7d26fd549a94f92f4a9d53b68245b996e/src/ducc0/sht/sht.cc#L1763-1791) we can see there is a distinct branch for geometry in `{"CC", "F1", "MW", "MWflip"}`  which I think is all the schemes with minimum `n_theta` ~ `L` rather than `2 * L`. This calls out to a function `resample_to_prepared_CC` which I suspect is doing something similar to the `upsample_by_two_mwss` calls within the `s2fft.forward` implementation. Appendix A in the paper [Reinecke, Belkner and Carron (2023)](https://arxiv.org/pdf/2304.10431) I think explains the double Fourier sphere technique being used here in more detail.

So if we want to support `n_theta = L + 1` in S2FFT I think we will need an upsampling / resampling implementation comparable to what we have for MW/MWSS and matching what is done within `ducc0` `resampling_to_prepared_CC scheme`.

matt-graham:
@jasonmcewen from my perspective this is now ready to merge. I have addressed the points discussed when I did the code walkthrough / review with you and Kevin, most significantly checking the quadrature rule tests for exact integration of polynomial functions tests the rules at the relevant minimum number of points, and also clarified the logic behind `m_offset` being equal to zero or one. If we want to support the $N_\theta \sim L$ grid dimensions for CC and F2 schemes that `ducc0` supports using the double Fourier sphere approach I'd say that would be best to do in a follow up PR.

mreineck:
> Appendix A in the paper [Reinecke, Belkner and Carron (2023)](https://arxiv.org/pdf/2304.10431) I think explains the double Fourier sphere technique being used here in more detail.

I can confirm that this is exactly how it is done. The beautiful aspect about this is that, after upsampling and applying the weights, you can downsample again and still carry out the Legendre transform part on the low number of rings.

PR #384, opened 2026-06-03
Use `ftm_to_flm` and `flm_to_ftm` primitives in each others transpose rules

matt-graham:
Resolves an issue I noted in https://github.com/astro-informatics/s2fft/pull/380#issuecomment-4578123508 that the `check_grads` gradient check utility was flagging incorrect order 2 gradients for the `ftm_to_flm` primitive when using reverse-mode automatic differentiation.

From a bit more digging and looking at the jaxprs of the `vjp` and `jvp` transforms applied to the primitives, I realised that while for forward-mode differentiation the jaxpr for the `jvp` of the primitives was being expressed in terms of the primitive itself, for reverse-mode differentiation the jaxpr for the `vjp` of the primitives was instead being expanded to the primitives used to define the (transpose) primitive rather than the transpose primitive itself. This means on subsequent applications of `vjp` (or `jvp`) we are differentiating through the underlying primitives, and due to the handling of the pole singularities this appears to introduce non-finite values.

This PR changes the transpose rules for the `ftm_to_flm` and `flm_to_ftm` primitives to instead be defined directly in terms of the other _primitive_ rather than the underlying `inverse_latitudinal_step_jax` and `forward_latitudinal_step_jax` implementations in the `otf_recursions` module. This means JVP and transpose operations over the pair are 'closed' in the sense that arbitrary combinations are expressed directly in terms of the primitives themselves so we can differentiate to any order without autodiffing through the underlying recursive implementations and hitting against issues with pole singularities. This also has the side benefit of removing the need to separately handle batching in the transpose rules.

I also added a note to explain why we are not using `is_linear=True` in `register_primitive` calls as this initially surprised me as primitives are linear in initial `flm` / `ftm` arguments. Having played around a little with trying to use both `deflinear` and `deflinear2` in `jax.interpreters.ad` to avoid explicitly having to define a JVP rule and to allow a simpler transpose definition, it doesn't look like this is easy to do while retaining ability to batch over `spin` and `precomps` arguments, so I have just added a note to remind us why this is the case to avoid similar confusion when looking at this again in future.

@ArtemBasyrov tagging you in case you can take a quick look over this as I think you're probably best placed to check this all makes sense and the change is quite small, but if you don't have capacity to look at this no worries obviously! If you're happy to do so I can add you as a collaborator on the repository so I can explicitly request a review.

### DIFF
--- a/s2fft/transforms/_ftm_flm_primitive.py
+++ b/s2fft/transforms/_ftm_flm_primitive.py
@@ -158,17 +158,10 @@ def _flm_to_ftm_jvp(primals, tangents, **params):
 
 
 def _flm_to_ftm_transpose(cotangent, flm, thetas, spin, *precomps, **params):
-    # ``flm`` arrives as an UndefinedPrimal; ``thetas``, ``spin`` and
-    # ``precomps`` are concrete residuals. The transpose of the inverse step
-    # is the forward step. We pass ``precomps=None`` so it regenerates the
-    # forward-direction precomps internally (the supplied ones are for the
-    # inverse direction).
-    def fn(c, s, _ignored_p):
-        return otf.forward_latitudinal_step_jax(
-            ftm_in=c, beta_in=thetas, spin=s, precomps=None, **params
-        )
-
-    cot_flm = _apply_with_batching(fn, cotangent, spin, precomps)
+    # The transpose of the flm_to_ftm primitive (applied to the initial flm argument) is
+    # the ftm_to_flm primitive. We do not pass through the supplied (flm_to_ftm) precomps
+    # so these are regenerated internally for the ftm_to_flm primitive.
+    cot_flm = _ftm_to_flm_primitive.bind(cotangent, thetas, spin, **params)
     return (cot_flm, None, None) + (None,) * len(precomps)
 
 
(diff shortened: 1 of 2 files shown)


s2fft/transforms/_ftm_flm_primitive.py
```diff
@@ -131,7 +131,15 @@ def _batch_primitive(primitive, batched_args, batch_axes, **params):
...
-    out_shape = flm.shape[:-_DATA_NDIM] + samples.ftm_shape(L, sampling, nside)
+    # As inverse_latitudinal_step_jax determines shape of first dimension of ftm value
+    # returned by size (of last dimension if batched) of thetas argument instead of
+    # the value returned by samples.ntheta / samples.ftm_shape(...)[0], to allow
+    # for the upsampling performed for MW + MWSS schemes, we cannot use
+    # samples.ftm_shape directly here as might be expected
+    out_shape = flm.shape[:-_DATA_NDIM] + (
+        thetas.shape[-1],
+        samples.ftm_shape(L, sampling, nside)[1],
+    )
```
matt-graham, line 142: This change was needed as we previously got an error when in tests of gradients of overall transforms due to shape of cotangent / primal values not matching expected shapes from abstract evaluation rules. Not quite sure why this wasn't causing an error previously though.

ArtemBasyrov:
Hi @matt-graham , these changes make sense to me. It's actually more closed-loop when defined through itself. As for the `is_linear=False`, you're right that the transform is only linear in respect to theta, but not spin. 

I'm also ok with being added as a collaborator for direct review requests.

PR #387, opened 2026-06-14
Fix HEALPix spin!=0 inverse/forward inaccuracy from Wigner-d node

slosar:
Note: this bug was uncovered and fixed by agentic AI tools, but it is a real non-trivial bug whose fix should be merged. We need sasfe_abs for avoit hitting nans for auto-diff. There is a test added that fails in the current main version, but passes here, so it would be good to keep it. Below is the machine generated description: 

The on-the-fly Price-McEwen Wigner-d recursion renormalises each step by bigi = 1/|dl_entry| and tracks lbig = log|dl_entry|. When an intermediate recursion value dl_entry is exactly zero (a node), bigi becomes inf and lbig becomes -inf, so dl_iter = inf*0 = NaN and the running log-norm goes to -inf. These NaNs are silently dropped by the nansum that accumulates ftm/flm, discarding that mode's contribution and producing percent-level errors.

Exact-zero nodes are essentially never hit by the generic theta samples of mw/mwss/dh/gl sampling, but HEALPix rings sit at rational cos(theta) values that land exactly on nodes for spin != 0, so spin-2 HEALPix transforms carried few-percent pointwise errors (spin-0 was unaffected and existing HEALPix tests only covered spin 0).

Guard the renormalisation so that where dl_entry == 0 we use bigi = 1 and lbig = 0, leaving lrenorm unchanged and setting the normalised value to 0, which is exactly consistent with the recursion bookkeeping. Applied to all four recursion paths (numpy/jax inverse and forward).

Adds a regression test comparing the recursive HEALPix spin-1/2 inverse against the independent Turok-recursion base transform.

### DIFF
--- a/s2fft/transforms/otf_recursions.py
+++ b/s2fft/transforms/otf_recursions.py
@@ -161,8 +161,16 @@ def inverse_latitudinal_step(
                     axis=-1,
                 )
 
-                bigi = 1.0 / abs(dl_entry[:, L_lower:])
-                lbig = np.log(abs(dl_entry[:, L_lower:]))
+                # Guard against renormalising on an exact node of the Wigner-d
+                # recursion (dl_entry == 0). This occurs at the rational cos(theta)
+                # values of HEALPix rings for spin != 0, where 1/|dl_entry| -> inf
+                # and log|dl_entry| -> -inf would otherwise inject NaNs that are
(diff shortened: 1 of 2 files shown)


slosar:
Can we merge this into main at some point, as our project now depends on a hacked s2fft to function properly?

